PR #5, opened 2021-01-22
Documentation

Gadgetoid:
Opened this for visibility - expands upon ede389d7af068197668ee85be2eaa0aa0985afc8 and starts to introduce API docs.

Still honing in on a style/voice here, but I have a strong preference for table-of-contents (make sure you have VSCode Markdown: All In one installed) and succinct examples/use cases to frame functions and types in context.

Some images would be really cool here, too,  specifically for the Pico Graphics library to illustrate things like clipping bounds, intersections, etc.

### DIFF
--- a/libraries/pico_display/README.md
+++ b/libraries/pico_display/README.md
@@ -69,7 +69,7 @@ int main() {
 
 ### PicoGraphics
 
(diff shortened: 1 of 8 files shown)


Gadgetoid:
As of #8 these docs will need a shakeup to match the new functions and camelcase types.

PR #12, opened 2021-01-23
Fixing C++ SDK documentation

jpwsutton:
The pico sdk documentation is missing the instructions for creating the PICO_SDK_PATH environment variable which cmake will then fail on later. This CR will add that as well as adds a section explaining how to install the toolchain for macos users.


Gadgetoid:
Thank you!

PR #15, opened 2021-01-23
Initial shot at is31fl3731

Gadgetoid:


Gadgetoid:
merged into `breakouts-dev` - https://github.com/pimoroni/pimoroni-pico/blob/breakouts-dev/drivers/is31fl3731/is31fl3731.cpp

PR #26, opened 2021-01-25
Change rectangle to rect to fix rectangle error on button press

JackBonnellDevelopment:
Also added random color on rect as default.

### DIFF
--- a/examples/pico_display/demo.cpp
+++ b/examples/pico_display/demo.cpp
@@ -22,7 +22,7 @@ void pixel(int x, int y, uint16_t c) {
   pico_display.frame_buffer[x + (y + 1) * 240] = c;
 }
 
-void rect(int x, int y, int w, int h, uint16_t c) {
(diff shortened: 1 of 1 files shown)


Gadgetoid:
As far as I can tell, this shouldn't work since `rect` was a type, that's now called `Rect` and the `rectangle` function is still:

https://github.com/pimoroni/pimoroni-pico/blob/032c17dd00dfcc49925afb66e8e622fef0ceb911/libraries/pico_graphics/pico_graphics.hpp#L79

JackBonnellDevelopment:
I receive this error after a clean pull on this repo 

`/home/jack/Development/pico/pimoroni-pico/examples/pico_display/demo.cpp:153:38: error: no matching function for call to pimoroni::PicoDisplay::rectangle(int, int, int, int)'
   pico_display.rectangle(0, 0, 18, 18);
                                      ^`

Gadgetoid:
Aha! Thank you. Yes it looks like that code was commented-out so the CI didn't bail on it and prompt us to fix it.

`rectangle` should take a `Rect(x, y, w, h)` like so: 

```c++
pico_display.rectangle(Rect(0, 0, 18, 18));
```

PR #28, opened 2021-01-26
Pico Unicorn Plasma

tkerby:
Added plasma example for Unicorn display based on the Adafruit display example from https://github.com/adafruit/RGB-matrix-Panel/blob/master/examples/plasma_32x32/plasma_32x32.ino

Modified size of plasma blobs and HSV conversion routines to be suitable for the Pico environment. Updated makefiles to build.

Note: It might be worthwhile by default diverting printf to USB as I have here in the other examples

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(pico_display)
 add_subdirectory(pico_unicorn)
+add_subdirectory(pico_unicorn_plasma)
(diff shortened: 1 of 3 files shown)


tkerby:
image

Gadgetoid:
This is really cool! Thank you.

You're probably right regarding UART. I've been using a PicoProbe for debugging since it doesn't bail my serial session every time I reload code- but we probably shouldn't make that assumption of everyone ese!

PR #32, opened 2021-01-27
Updated MicroPython Pico Display documentation

UnfinishedStuff:
I've expanded the MicroPython Pico Display documentation to cover all functions, trying to keep in the style of what was currently there.  

I've also included the clips diagram from my documentation repo: @Gadgetoid I know you said you wanted to incorporate some sort of image but I wasn't sure exactly what you had in mind, so I've put that in as a placeholder until something better can be put together.

Happy to change it as required.


Gadgetoid:
This is already *so* far beyond what I'd ever expect from a PR that I'm happy to merge away and fix stuff ad-hoc. But some observations:

1. We should probably observe PEP8 codestyle for better or worse (I might not have!), in particular function arguments should be `function(x, y, z)` rather than `function(x,y,z)`.... pedantry ahoy!

2. Where something like `character` is documented, a quick inline example of getting an individual char code might be handy. IE: `char_a = ord('a')`

For graphics I was hoping to shore up the Pico Graphics library just enough to make it output images so we could do really accurate "in-engine" demos of every drawing operation (pixel doubled if necessary). If this sounds like an over-ambitious bout of pure madness that we almost certainly don't have time for... that's probably because it is. This does mean that code usage example and *actual* output can sit side-by-side though.

Gadgetoid:
Okay re graphics demos I had an idea and apparently 99% of the work is already done for me so I only had to glue Pico Graphics into 32blit and come up with...

image

Gadgetoid:
And with not too much more tinkering, here's what I was going for:

image

image

Gadgetoid:
Who knows how useful *this* is though:
image

UnfinishedStuff:
OK, I've taken a read through of PEP8 and I _think_ I've caught everything?  And I've doc'd the `char_a` example, I'd entirely forgotten that `ord()` was a thing.  

I've also nuked my clip diagram to make space for your much prettier diagrams.  I love the style, my only comment would be that they'll be quite a chunky interruption to the text.  I think this is probably necessary for the `clip` function.  I tinkered with the below for `rectangle`, but the origin pixel gets a bit lost.  What do you think?

altDiag

Gadgetoid:
Maybe I'm just getting old, but I *like* chunky interruptions to text. Text feels a little overwhelming sometimes, specially when it's as information dense and mentally heavy as this. Then again I... don't really read documentation! The *huge* images (ignoring them being the uh... resolution that 32blit happens to output and I was being lazy) feel like a rest stop for the brain.

You're wholly right about the "origin" pixel getting a little lost, and an "x/y" annotation is a lot clearer. Can we meet in the middle? :laughing:

UnfinishedStuff:
Oh absolutely, the size thing was just a suggestion.  I wasn't sure of the best way to store the images without cluttering the directory, so I create an `images` folder in the Pico Display documentation directory, uploaded them there, and inserted the images in-line with the text.  

What do you think?  Would you prefer the images in the same directory as the readme?

Gadgetoid:
Sorry been tackling household stuff for the last week or so! I've merged this because *it's fantastic* and the more viability it gets the better. Thank you!

PR #33, opened 2021-01-28
Fix typo on `vaule`

rogerthomas84:


Gadgetoid:
Good catch, thanks for taking the time to fix it!

PR #39, opened 2021-02-03
RV3028 RTC Driver and example for Raspberry Pico

simon3270:
I've coded up the RV3028 driver for the Pico, so that I could use you RTC breakout. I started with the Arduino RV_3028_C7 library (version 2.1.0), added startup code, and modified the register reads and writes to use the same code as the existing drivers (particularly the MSA301). I also added the method to get the current timer countdown count (missing from the Arduino code, though the constants to define the register were present).

I include C/C++ example code to test the driver. This requires a display and the RTC module - I used your pico display and wired the RTC module to GPIO pins 20, 21 and 22 (like in the Pi400 Breakout Garden). I haven't tested the alarm code yet, though the timer code seems to work fine. My example also doesn't contain the code to set the time - you would need to do that on another machine.

There is no documentation on the calls (like the other drivers that are being developed!), and none in the original Arduino code, as far as I can tell. I worked out the methods from the example code given, and the RV3028C7 datasheet. Maybe that's something for a later date....


Gadgetoid:
Builds and runs great for me- thank you! Was also really easy to port to Pico Explorer. I wonder if it's worth switching the demo over to that, or having some `#ifdef` setup.

simon3270:
Thanks! Yes, #ifdef should work (I was pondering that, but couldn't decide which should be the default!). Would it be possible to have a single header file for the explorer and display, which worked out which was being used? It would have to cope with things like set_led() only being on the display.

I have spotted one bug in the driver (getTimerCount() effectively ignores the top byte), so I'll fix that.

I'll also add some code I've done to make the + and - buttons in the demo autorepeat. Not having that was probably why I didn't test the getTimerCount fully - pressing the "+" button 256+ times would take a while...

PR #46, opened 2021-02-08
RV3028 driver bug, autorepeat on example buttons

simon3270:
Bug in getTimerCount() in driver (was ignoring top byte).
Add auto-repeat on the +/- timer buttons in the example.
Allow PicoExplorer/PicoDisplay selection in the example with #define USE_PICO_EXPLORER 1.
Note that CMakeLists.txt includes the display and explorer libraries, but only one is used.

### DIFF
--- a/drivers/rv3028/rv3028.cpp
+++ b/drivers/rv3028/rv3028.cpp
@@ -528,7 +528,7 @@ namespace pimoroni {
   {
     // Reads the number of remaining timer ticks
     uint8_t r0 =readRegister(RV3028_TIMERSTAT_0);
-    return(r0 + (readRegister(RV3028_TIMERSTAT_1) >> 8));
+    return(r0 + (readRegister(RV3028_TIMERSTAT_1) << 8));
   }
 
   void RV3028::enableTimerInterrupt()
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Good grief, you're fast!

I think we should build repeat into the libraries somehow. Right now I just throw my AutoRepeat class at everything:

https://github.com/Gadgetoid/pico-rgb-keypad-debounce/blob/b609f525b067941152c657d2f98355bb33d2fbe8/main.cpp#L11-L57

```c++
AutoRepeat autorepeat_a;

while(true) {
    auto a_pressed = autorepeat_a.next(pico_display.is_pressed(pico_display.A);
}
```

Which dramatically simplifies the "main" loop.

simon3270:
I was helped by having commented the Display->Explorer transform in the code, so just had to follow my own instructions!
(That, and I have been doing this for 40 years, so have had a bit of practice!).  I'd also just finished a very boring work Zoom meeting, so needed some mental stimulation :-)

For the autorepeat, I wanted to leave the keys on the left as single-shot (to select what the clock was going to do), and the ones on the right mostly repeaters, but with a varying repeat speed (gets faster after the first 10 increment/decrements). I did start writing code to use the alarm setting functions, but got bogged down in which key did what, and which bits of information you'd have to select (there seem to be 8 different alarm modes). I was also getting to the point where I needed short and long presses, separately from autorepeat, so it was all getting a bit complicated. Probably need to leave that for another day.

Gadgetoid:
Works a treat on my lazy Pico Explorer setup. Got a piezo on there too :thinking: 

Thank you!

PR #47, opened 2021-02-08
Added Pico Scroll documentation

UnfinishedStuff:
I've added a MicroPython documentation page for the Pico Scroll Pack.  I've tried to keep it in the same style as the Display Pack docs and the docs for the other scroll boards.  Happy to tweak as needed.

### DIFF
--- a/micropython/modules/pico_scroll/README.md
+++ b/micropython/modules/pico_scroll/README.md
@@ -1,0 +1,98 @@
+# Pico Scroll Pack - MicroPython <!-- omit in toc -->
+
+Our Pico Scroll Pack offers a 17x7 white LED matrix for your Raspberry Pi Pico.  It also includes four buttons!
+
+We've included helper functions to handle every aspect of drawing to the matrix and interfacing with the buttons. See the [function reference](#function-reference) for details.
+
+- [Example Program](#example-program)
+- [Function Reference](#function-reference)
+  - [init](#init)
+  - [get_width](#get_width)
+  - [get_height](#get_height)
+  - [set_pixel](#set_pixel)
+  - [update](#update)
+  - [clear](#clear)
+  - [is_pressed](#is_pressed)
+
+## Example Program
+
+The following example sets up the matrix, sets each pixel to an increasing brightnesses level, and then clears the matrix only after button A is pressed.
+
+```python
+import picoscroll
+
+# Initialise the board
+picoscroll.init()
+
+brightness = 0
+
+# For each pixel in the matrix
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Awesome, thank you! Better to have this visible for everyone ASAP so I've merged. We're currently evaluating how we present documentation and @ZodiusInfuser is working on way to generate a nice structured documentation website for this stuff.

ZodiusInfuser:
Indeed, so @UnfinishedStuff if there's anything particular you think we should try and do, then let us know either on here or via other comms (I'm @ZodiusInfuser on most things). The intention at the moment is to stick with markdown for the pages, so we can either move or duplicate the documentation that's currently in this repo.

UnfinishedStuff:
@ZodiusInfuser I think the structure and layout of the documentation is grand really, the one thing I'd try to address at the moment is the visibility of it.  There have been a few people on the Pimoroni forum in the past week asking if there was any documentation because they couldn't find it.  Maybe the shop pages for each product should have a link to the documentation as well as the library (in a similar vein to Adafruit's product pages), or a link somewhere on the main `/pimoroni-pico/README.md` page?  Though that might get cluttered quickly.

Something else which might be useful is a clearer explanation of the fact that all of the libraries are bundled into the custom Pimoroni firmware.  There have been a few cases of people not realising that the custom Pimoroni firmware needs to be installed rather than just using the official MicroPython firmware.  A few others haven't been sure how to copy it over: I think they were more familiar with CircuitPython modules which can be copied and pasted over in USB mode, rather than being installed as part of a bundled firmware, or the standard Pi where Python is already installed on the board and you install the module beside it.  The product pages talk about "libraries", and I don't think many people realises these are bundled in to the custom firmware, so maybe make that a bit more explicit?  Maybe something like "C/C++ and MicroPython libraries are available at [x], and the MicroPython libraries have been bundled up into a custom firmware for easy installation."?

helgibbons:
Thanks for the useful feedback @UnfinishedStuff - I've updated the shop pages to state specifically that the MicroPython image has our libraries included, and we'll have a think about how we can incorporate more direct links from the shop pages as the docs become more settled.

I've tried to catch the most common beginner questions in the Getting Started tutorial (at least up to the point of running the examples) and I'm updating the troubleshooting section as I come across frequently encountered problems - https://learn.pimoroni.com/tutorial/hel/getting-started-with-pico . If you have any suggestions for stuff that needs adding do let me know! 

Thanks again for all your contributions to the docs, they are much appreciated :)

PR #48, opened 2021-02-09
Updated demo.py with A/B/X/Y button example

SimUKdev:
Added code to show a simple example of the on-board A/B/X/Y button press detection and displaying the result on the display, as this feature of the Pico Explorer hardware was not being demonstrated in the initial demo.py example.


Gadgetoid:
Great idea, thank you!

PR #55, opened 2021-02-11
Adding support for differing font heights

Pinkyatacme:
As the pico display pack means we can do taller (and more readable) fonts, I've adjusted the character, text and font code to allow up to 6x8 fonts. Wider fonts may follow if I get the itch to implement that.


Pinkyatacme:
Updated to add example 6x8 font.

Gadgetoid:
Sorry I've been hesitant to merge this since I thought we had markdown-fuelled text rendering in the pipeline, but it hasn't been forthcoming and it seems a shame to leave this floundering! Will try to get it tested & merged today and we can go from there. (a big ugly rebase and merge in my future I suspect)

Gadgetoid:
I think there should be a better way to switch between fonts on a per project basis but I'm far from the authority on this.

I've come up with https://github.com/pimoroni/pimoroni-pico/commit/165605b21780569f77bc7c6b4136e7cd96b32f6e which may or may not work, but illustrates what I'm going for. Rather than renaming a file, any single project can just "include" the front file it needs, and use `set_font` to enable it.

Gadgetoid:
Fixed a bunch more stuff including docs, rebased onto main to avoid some of the messy commits here and raised #70. Fingers crossed the CI passes and things still work because I'm in over my head here. Continue discussion at #70.

PR #65, opened 2021-02-21
Explorer tga order fix

jpwsutton:
Was tearing my hair out as to why my personal TGA files were the wrong colour, then after a bit of digging I realised that the colour order isn't RGB, it's BGR!

http://www.paulbourke.net/dataformats/tga/


Gadgetoid:
Haha! good catch. The rainbow fox didn't make it quite so obvious. Thank you.

PR #66, opened 2021-02-22
Pico Driver for vl53l1x Breakout

simon3270:
Another driver - I had a VL53L1X Time of Flight breakout that I thought the Pico could use.
As I did for the rv3028 driver, I took Arduino code and modified the startup and the interface to the actual device, and used the Pico timing code.
As far as I can tell, this code is under BSD license, but still allows modification. I've left the original copyright text in place, in the LICENSE file.
The demo example program puts a tape measure on the Pico Display or Pico Explorer (set a define to get the Explorer version). Three buttons select Metric/Imperial, set the mode (Short, Medium or Long range), and hold the display (for those times you can't read the display while actually measuring). I'm reasonably happy with the demo, though you do have to press the buttons for tenths of seconds before they register (none of them auto-repeat, so pressing too long is not a problem). While getting this working I discovered the enormous range of capabilities for this apparently simple device. My demo doesn't stress those functions, but I fully expect them to work.


simon3270:
Hmm, just noticed the other commits that came along with mine. I thought I'd started with a clean copy of "main", but clearly not. Sorry!

Gadgetoid:
You should be able to do something like `git fetch origin main` `git rebase origin/main` to apply your last commit to a clean version of upstream. Assuming you have `origin` set up as this repo, or you can add an upstream for this purpose: `git remote add upstream https://github.com/pimoroni/pimoroni-pico` and `git fetch upstream main` etc.

TLDR: rebase your branch, it *might* go smoothly.

Alternatively you could build a fresh branch and `cherry-pick` this commit from it. Happy to help (though I'm just about to dash out for a walk) with the appropriate git-fu to resolve this.

Also- thanks for whipping up a driver! I might even be able to test this :laughing:

Gadgetoid:
Since the commit history was a bit of a mess, I managed to produce: https://github.com/pimoroni/pimoroni-pico/compare/driver-vl53l1x

With:
```
git remote add upstream https://github.com/pimoroni/pimoroni-pico
git checkout -b new-branch-name upstream/main
git cherry-pick 9e6cf8b97785fa4e10c8bba1402637909a918343
... resolve the merge conflict in CMakeLists.txt by removing the merge stuff and the pio subdir
git add drivers/CMakeLists.txt
git rebase --continue
```

PR #68, opened 2021-02-23
Pico driver for vl53l1x Time of Flight breakout

simon3270:
As for the rv3028 driver, this VL53L1X driver is ported from the corresponding Arduino driver. It's apparently BSD 3-Clause licensed (LICENSE file included). The driver seems to function well, though I haven't explored the minutiae of modifying angles of reception or of sensor timing (beyond the default values used for the 3 range modes).
The demo pico_tof_display acts as a tape measure. Button X selects Metric/Imperial for the units used to display the distance, Y selects the range mode (Short, Medium or Long), and B holds the reading (in case you can't see the display while measuring). The buttons seem sluggish (have to be held for longer than I saw on the rv3028 driver), but none auto-repeat so pressing for too long is not a problem.
Many thanks to @Gadgetoid for sorting out my git problems, and cherry-picking just this fix (10 years of using git and I still get lost!).


Gadgetoid:
I still get lost. The cherry pick was a last desperate attempt because I couldn't cleanly rebase :grimacing: 

Thanks for writing this, about to give it a shot!

Gadgetoid:
Added a commit with some (minor, honest!) tweaks. Sensor output seems sensible to me, and non-blocking mode makes it easier to handle button input. Though I suspect there should be some sort of interrupt driven read here where we can always know if a new sample is available and read it.

simon3270:
The tweaks may be minor, but they are great! I As you say, the buttons are now much more responsive, and I do like the change for the hold button (in my code, you had to hold on to the button to freeze the display - I prefer your "tap to freeze, tap again to release").

Gadgetoid:
Great! Thank you for pulling this together- much appreciated!

ZodiusInfuser:
@simon3270 Thanks for your work on this and the rv3028. In case you have not seen, I have set up a dedicated branch for doing breakout development: https://github.com/pimoroni/pimoroni-pico/tree/breakouts-dev

In it I have created boilerplate code for all the breakout drivers, which we are slowly going through. So if you would like to contribute, that would be the best branch to pull from and PR to going forward.

I have already merged your two drivers in, and updated their coding style to what we are using. Not saying it's the best style, but it is what it is. Hopefully I did not break anything in the process.

PR #70, opened 2021-02-23
Adding support for differing font heights (Replaces #55)

Gadgetoid:
This includes the variable font support by @Pinkyatacme raised in #55, rebased onto main to eliminate some redundant commits and enhanced with the ability to `set_font` at runtime. I've added an additional demo to showcase this.

This PR should replace #55.


Gadgetoid:
Looks like there might be issues with text rendering (and/or underlying font data) in both versions of the font:

Image from iOS

Pinkyatacme:
Sorry about that - PR #71 adds the right version. Always a risk of these errors when hand-coding the font from graph paper, and I somehow checked in the 1st attempt!

Gadgetoid:
You....

hand-coded from graph paper?

:scream: 

Truly the real O.G. way to do it. Respect!

Lovely font, by the way!

Gadgetoid:
Thanks - I've cherry picked your commit over and merged it into the renamed font file, preserving your authorship.

You should probably grab this branch and modify from here so we're working on the same thing.

Looks like that fixes the majority of characters, any idea what's happened to £$%... though actually £ is non-ascii which would explain why it horribly breaks in both sets. Cursed ASCII :laughing: I'd better fix that!

Image from iOS (1)

Gadgetoid:
yup non-ASCII £ did it. Was probably ending up as unicode and getting treated as two chars indexed off the end of the array in Mystery Memory, here we go:

Image from iOS (2)

PR #75, opened 2021-02-25
Support portrait orientation using new constructor with width, height values

hamid-elaosta:
Small change to add support for portrait orientation by accept width/height in a new constructor. Using a new constructor stops it from breaking existing examples/code.

It currently only works for the orientation of the display board (keys right-way-up), an "orientation" flag would need to be added for upside-down.

```
PicoDisplay pico_display(buffer, PicoDisplay::PORTRAIT_WIDTH, PicoDisplay::PORTRAIT_HEIGHT);
```


Gadgetoid:
Nice work! Just tried it out here and it's working well.

I'm not averse to having an "orientation" or "rotation" flag, but in that case the `WIDTH` and `HEIGHT` should probably always be supplied in the same order, and rotation does two things:

1. Sets `MADCTL` (best register name ever) to configure rotation
2. Determines how the supplied width/height are stored and allows you to use `pico_display.width` and `pico_display.height` to get the buffer width/height for drawing (these values are currently stored on the ST7789 driver but private so not user accessible).

In fact in this case your extra constructor would *just* take a `rotation` argument, and use the internal width/height accordingly. I think that can work?

I suspect the ST7789 driver should, ideally, not hardcode any width/height values either but that's probably a TODO. :laughing: 

I was going to suggest modifying the Pico Display demo to showcase both Portrait and Landscape orientations... but it, uh, turns out the demo is a hot mess at the moment.

Gadgetoid:
Currently working on #103 in lieu of this.

hamid-elaosta:
Apologies for not being able to follow up on that PR. Family life has meant I haven't been doing any development for a bit. Looking forward to the updated lib, I'll be sure to update my menu library once this is available!

Gadgetoid:
No worries!

I think you spurred us on to make some changes to ST7789, and just to be awkward there's now support for the *round* LCD in there too. Thank you!

PR #81, opened 2021-03-02
driver-sh1170 for Raspberry Pico (1.12" OLED I2C Breakout)

simon3270:
Needed a display to attach to a Tiny 2040, so I ported one for the OLED 1.12" monochrome I2C Breakout.
As before, I found an Arduino driver and modified it to the new breakouts-dev standards and converted from C to C++.
I only have the I2C version of the breakout, so I couldn't check the SPI code.
As far as I remember, the Python version of the Pi library (luma) has an image display function, but that wasn't in the original code I used. I'm going to have to dig around for that!
There is a warning in the demo code about converting a string constant to a char *. I also cast an int to an "enum" type - not sure how to use the enum directly (for black/white colours).
I have created this PR against the breakouts-dev branch - I hope that's right! Thanks for setting up all of the breakout templates - made life a lot easier.


Gadgetoid:
I'd got about as far as burning myself on my 1.12" OLED (an old prototype held together with tape!) and you sweep in with this. Well timed, thank you! It's on our radar.

Gadgetoid:
Thank you- I'm going to pull this code apart somewhat to make constructors for i2c and spi separately, since we wont be able to support i2c/spi in Python using preprocessor macros. Will push it up to a branch here for comparison.

Gadgetoid:
Raised a PR of my changes so far against your branch, so you can see what I'm changing - https://github.com/simon3270/pimoroni-pico/pull/1/files

Right now I'm successfully driving i2c and SPI displays side-by-side with the same code and a shared display buffer.

Your `char *` issue was solved by switching to `const char*`

I attempted to use an `enum class SH1107_COLOR: uint8_t` to fix your enum woes, but it's only made things more complex. If less wordy. I think just using `bool` and `#define SH1107_BLACK false` and `define SH1107_WHITE true` would work here... although black/white are relative if you've inverted the display anyway so maybe it should be `SET` and `CLEAR`.

ZodiusInfuser:
@Gadgetoid I see you're making tweaks for SPI and I2C, so let me know when it's good for me to merge this PR 🙂

SamStudio8:
Hey @Gadgetoid, I know you've got a lot going on, I was just wondering on the status of this PR -- is it still on the radar?

Gadgetoid:
@SamStudio8 yes, I'm not sure why it's still in limbo. Probably needs a thorough code review and a look at Python bindings.

Have you got it up and running at all?

SamStudio8:
I hadn't tried but perhaps this is an excuse to learn how to build an image :thinking:

Gadgetoid:
Oh I don’t think there are any MicroPython bindings yet (so this would only work in C++)- but if this PR weren’t so ancient you could get a build right from the continuous integration.

Gadgetoid:
This has fallen by the wayside, but I'd like to try and refactor and include it at some point.

These notes are mainly to my future self.

Key points:

* [ ] Should prefer arguments passed to the constructor rather than compile-time config (for MicroPython support)
* [ ] Should use bitmap_fonts and hershey_fonts libraries instead of baking its own fonts (those probably need some compile time defines for configuring included fonts)
* [ ] Should use Pico Graphics if possible, I'm trying to canonicalise on a single drawing API

I'm aiming at something like this which combines the raw ST7789 driver with Pico Graphics -  https://github.com/pimoroni/pimoroni-pico/pull/327

Gadgetoid:
I've been rewriting much of how we handle graphics & display drivers here - https://github.com/pimoroni/pimoroni-pico/pull/373

As such this driver would need to inherit from `PicoGraphics::DisplayDriver` and implement `void update(PicoGraphics *display)` method.

The ST7789 and ST7735 drivers both do this, eg:

```c++
  // Native 16-bit framebuffer update
  void ST7735::update(PicoGraphics *graphics) {
    if(graphics->pen_type == PicoGraphics::PEN_RGB565) {
      command(reg::RAMWR, width * height * sizeof(uint16_t), (const char*)graphics->frame_buffer);
    } else {
      command(reg::RAMWR);
      gpio_put(dc, 1); // data mode
      gpio_put(cs, 0);

      graphics->scanline_convert(PicoGraphics::PEN_RGB565, [this](void *data, size_t length) {
        spi_write_blocking(spi, (const uint8_t*)data, length);
      });

      gpio_put(cs, 1);
    }
  }
```
  
In the case of SH1170, these OLED displays are monochrome so we'd need a 1-bit pen type added to PicoDisplay. This is on my radar at least.

Standardizing on Pico Graphics should give us monochrome dithered jpeg support, plus our common Bitmap and Hershey fonts on these OLEDs.

Gadgetoid:
Due to a pressing need to use this OLED for whacky testing purposes we integrated support into PicoGraphics yesterday - https://github.com/pimoroni/pimoroni-pico/pull/373

Thank you for this working reference implementation, I think we'd have lost a lot more hair without it!

Our driver currently only supports 128x128 displays with normal rotation, but that can be improved over time!

Example Python code:

```python
from picographics import PicoGraphics, DISPLAY_I2C_OLED_128X128, PEN_1BIT
from pimoroni_i2c import PimoroniI2C

bus = PimoroniI2C(4, 5)

display = PicoGraphics(DISPLAY_I2C_OLED_128X128, pen_type=PEN_1BIT, bus=bus)

display.set_pen(1)
display.clear()
display.set_font("gothic")
display.set_pen(0)
display.text("Hello World", 0, 10, 0, 0.5)
display.update()

```

simon3270:
Glad to have been of service!

Gadgetoid:
Closing this now since the Pico Graphics based version (albeit very incomplete) obsoletes it.

Thank you!

Issue #80, opened 2021-03-02
Suggestion: Screen orientation setting


Galbi3000:
The picodisplay seems to be fixed orientation so that the "Pico Display" logo on the silkscreen is on the left. But the way I need to use it due to the USB cable I need the logo on the right so now everything is upside-down.

On the Arduino display libraries, either by Adafruit or other compatible libraries, there is a function to rotate the display so any of the edges are the top edge so you can have it 0,90, 180 or 270 degrees rotated.

Any chance a similar function could be added to the display library and module here?

Pinkyatacme:
A minor adjustment to the st7789 driver code inverts the display (on line 59 change "\x70" to "\xB0". As there's only a minor adjustment needed to flip the display in both landscape and portrait formats (inverting bits 6 & 7 of the MADCTL value), it should ab fairly simple to add an additional parameter to the display.

hamid-elaosta:
Check out my PR from about a week ago if you want a quick way to change orientation. https://github.com/pimoroni/pimoroni-pico/pull/75

Pinkyatacme:
I have a local version that implements this (thanks for the pointers Hamid), but it's not exactly "correct". On looking at the code, doing it properly is a little more complex, as the 'command' function expects a string (that is then converted into a uint8!), so a simple XOR of the MADCTL value isn't quite as simple as I had hoped.

hamid-elaosta:
This library from Adafruit has an orientation function, perhaps it could be used for implementation pointers?

https://github.com/adafruit/Adafruit-ST7735-Library/blob/master/Adafruit_ST77xx.cpp

Pinkyatacme:
Just added PR #85 to implement a flip command - a call of pico_display,flip() [assuming that's your display object name!] should rotate the screen 180. I would have preferred to use the RDDMADCTL call to fetch the current data, but struggled to get that working (it seems to need a dummy spi read to get to the data, and I can't seem to figure that one out).

Galbi3000:
> Just added PR #85 to implement a flip command - a call of pico_display,flip() [assuming that's your display object name!] should rotate the screen 180. I would have preferred to use the RDDMADCTL call to fetch the current data, but struggled to get that working (it seems to need a dummy spi read to get to the data, and I can't seem to figure that one out).

That would solve my issue perfectly. I don't have the means to compile the source so I will have to wait for it to be added to the main branch and the firmware image file is updated.

Pinkyatacme:
[...]

PR #85, opened 2021-03-06
Slight refactor of st7789 driver, and addition of the flip function

Pinkyatacme:
As per issue #80, added a flip function. Required minor refactor of code to create a madctl variable for the st7789, as trying to read and update the registers was a miserable failure.

### DIFF
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -51,11 +51,11 @@ namespace pimoroni {
       sleep_ms(150);
 
       if(width == 240 && height == 240) {
-        madctl = 0x04;
+        madctl[0] = 0x04;
       }
 
       if(width == 240 && height == 135) {
-        madctl = 0x70;
+        madctl[0] = 0x70;
       }
       
       command(reg::MADCTL,    1, madctl);  // row/column addressing order - rgb pixel order
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -58,6 +58,10 @@ namespace pimoroni {
         madctl[0] = 0x70;
       }
       
+      if(width == 135 && height == 240) {
+		madctl[0] = 0x00;
+      }
+
       command(reg::MADCTL,    1, madctl);  // row/column addressing order - rgb pixel order
       command(reg::TEON,      1, "\x00");  // enable frame sync signal if used
       command(reg::COLMOD,    1, "\x05");  // 16 bits per pixel
@@ -77,6 +81,11 @@ namespace pimoroni {
       if(width == 240 && height == 135) {
         command(reg::RASET,     4, "\x00\x35\x00\xbb"); // 53 .. 187 (135 rows)
         command(reg::CASET,     4, "\x00\x28\x01\x17"); // 40 .. 279 (240 columns)
+      }
+
+      if(width == 135 && height == 240) {
+        command(reg::CASET,     4, "\x00\x34\x00\xba"); // 52 .. 186 (135 rows)
+        command(reg::RASET,     4, "\x00\x28\x01\x17"); // 40 .. 279 (240 columns)
       }
     }
 
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -7,6 +7,7 @@
 #include "hardware/pwm.h"
 
 namespace pimoroni {
+  char madctl[2];
 
   void ST7789::init(bool auto_init_sequence) {
     // configure spi interface and pins
@@ -50,15 +51,16 @@ namespace pimoroni {
       sleep_ms(150);
 
       if(width == 240 && height == 240) {
-        command(reg::MADCTL,    1, "\x04");  // row/column addressing order - rgb pixel order
-        command(reg::TEON,      1, "\x00");  // enable frame sync signal if used
-        command(reg::COLMOD,    1, "\x05");  // 16 bits per pixel
+        madctl = 0x04;
       }
 
       if(width == 240 && height == 135) {
-        command(reg::MADCTL,    1, "\x70");
-        command(reg::COLMOD,    1, "\x05");
+        madctl = 0x70;
       }
+      
+      command(reg::MADCTL,    1, madctl);  // row/column addressing order - rgb pixel order
+      command(reg::TEON,      1, "\x00");  // enable frame sync signal if used
+      command(reg::COLMOD,    1, "\x05");  // 16 bits per pixel
 
       command(reg::INVON);   // set inversion mode
       command(reg::SLPOUT);  // leave sleep mode
@@ -146,5 +148,10 @@ namespace pimoroni {
 
   void ST7789::vsync_callback(gpio_irq_callback_t callback) {
     gpio_set_irq_enabled_with_callback(vsync, GPIO_IRQ_EDGE_RISE, true, callback);
+  }
+
+  void ST7789::flip(){
+    madctl[0] ^= 0xC0;
+    command(reg::MADCTL,    1, madctl);
   }
 }
--- a/drivers/st7789/st7789.hpp
+++ b/drivers/st7789/st7789.hpp
@@ -47,6 +47,7 @@ namespace pimoroni {
     void vsync_callback(gpio_irq_callback_t callback);
     void update(bool dont_block = false);
     void set_backlight(uint8_t brightness);
+    void flip();
 
     enum reg {
       SWRESET   = 0x01,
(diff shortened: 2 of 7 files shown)


Pinkyatacme:
Apologies for the 2 borken commits. Note to self - stop working outside the tree!

Pinkyatacme:
Just for completeness, I've retrofitted the portrait mode (pr #75) with the flip screen function.

Pinkyatacme:
Following a comment on the issue note, I've also added the Python updates.

PS. Thanks for the well documented updates to the Micropython repo - made it easy to get the firmware updated and tested :D

Gadgetoid:
As near as I can tell you never set or transfer `madctl[1]`, could `madctl` just be a `uint8_t`?

This goes... weird... on the 240x240 1.3" LCD:

Calling:
```
        screen.init();
        screen.flip();
```

Results in:

image

And unflipped the image is:

image

I think MADCTL can also be set entirely during `init()` by adding some `v_flip`, `h_flip` and (IIRC) `xy_swap` arguments or even just- as per #75- a "rotation" argument.

Yes I'm being awkward by testing this changeset with our 240x240 breakout :laughing:

Pinkyatacme:
Good thinking to test on the 240x240!
The reason for madctl being an array is the command function would also need to be recreated. I did look at what would be needed, and it was a lot more than simply having an array with a redundant element. The way that the MADCTL works is rather odd at first sight, but consistent with what I've seen in other display/graphic controllers - one bit mirrors X, one mirrors Y. I suspect that the base MADCTL is a little off when using 240x240 - back to the datasheet I go! ;-)

Gadgetoid:
Oh just to add fuel to the fire I also have a *round* display here :eyes: 

I decided to collate all the particulars of `MADCTL` into an enum to handle it better:

```patch
diff --git a/drivers/st7789/st7789.hpp b/drivers/st7789/st7789.hpp
index 76bef36..4b352ad 100644
--- a/drivers/st7789/st7789.hpp
+++ b/drivers/st7789/st7789.hpp
@@ -21,7 +21,7 @@ namespace pimoroni {
     int8_t sck    = 18;
     int8_t mosi   = 19;
     int8_t miso   = -1; // we generally don't use this pin
-    int8_t bl     = 20;
+    int8_t bl     = 21;
     int8_t vsync  = -1; // only available on some products
 
     uint32_t spi_baud = 64 * 1024 * 1024;
@@ -41,7 +41,7 @@ namespace pimoroni {
       width(width), height(height),      
       cs(cs), dc(dc), sck(sck), mosi(mosi), miso(miso), frame_buffer(frame_buffer) {}
 
-    void init(bool auto_init_sequence = true);
+    void init(bool auto_init_sequence = true, bool round = false);
 
     void command(uint8_t command, size_t len = 0, const char *data = NULL);
     void vsync_callback(gpio_irq_callback_t callback);
@@ -49,6 +49,20 @@ namespace pimoroni {
     void set_backlight(uint8_t brightness);
     void flip();
 
+    enum MADCTL : uint8_t {
+      ROW_ORDER   = 0b10000000,
+      COL_ORDER   = 0b01000000,
+      SWAP_XY     = 0b00100000,  // AKA "MV"
+      SCAN_ORDER  = 0b00010000,
+      RGB         = 0b00001000,
+      HORIZ_ORDER = 0b00000100
+    };
+
+    #define ROT_240_240_0      0
+    #define ROT_240_240_90     MADCTL::SWAP_XY | MADCTL::HORIZ_ORDER | MADCTL::COL_ORDER
+    #define ROT_240_240_180    MADCTL::SCAN_ORDER | MADCTL::HORIZ_ORDER | MADCTL::COL_ORDER | MADCTL::ROW_ORDER
+    #define ROT_240_240_270    MADCTL::SWAP_XY | MADCTL::HORIZ_ORDER | MADCTL::ROW_ORDER
+
     enum reg {
       SWRESET   = 0x01,
       TEON      = 0x35,

```

And then store as `uint8_t` and handle conversion to `char *` when it's transmitted:

```
diff --git a/drivers/st7789/st7789.cpp b/drivers/st7789/st7789.cpp
index 0c1c567..705641f 100644
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -7,9 +7,11 @@
 #include "hardware/pwm.h"
 
 namespace pimoroni {
-  char madctl[2];
+  uint8_t madctl;
+  uint16_t caset[2] = {0, 0};
+  uint16_t raset[2] = {0, 0};
 
-  void ST7789::init(bool auto_init_sequence) {
+  void ST7789::init(bool auto_init_sequence, bool round) {
     // configure spi interface and pins
     spi_init(spi, spi_baud);
 
@@ -50,19 +52,6 @@ namespace pimoroni {
 
       sleep_ms(150);
 
-      if(width == 240 && height == 240) {
-        madctl[0] = 0x04;
-      }
-
-      if(width == 240 && height == 135) {
-        madctl[0] = 0x70;
-      }
-      
-      if(width == 135 && height == 240) {
-               madctl[0] = 0x00;
-      }
-
-      command(reg::MADCTL,    1, madctl);  // row/column addressing order - rgb pixel order
       command(reg::TEON,      1, "\x00");  // enable frame sync signal if used
       command(reg::COLMOD,    1, "\x05");  // 16 bits per pixel
 
@@ -74,19 +63,38 @@ namespace pimoroni {
 
       // setup correct addressing window
       if(width == 240 && height == 240) {
-        command(reg::CASET,     4, "\x00\x00\x00\xef");  // 0 .. 239 columns
-        command(reg::RASET,     4, "\x00\x00\x00\xef");  // 0 .. 239 rows
+        caset[0] = 0;
+        caset[1] = 239;
+        raset[0] = round ? 40 : 0;
+        raset[1] = round ? 279 : 239;
+        madctl = MADCTL::HORIZ_ORDER;
       }
 
       if(width == 240 && height == 135) {
-        command(reg::RASET,     4, "\x00\x35\x00\xbb"); // 53 .. 187 (135 rows)
-        command(reg::CASET,     4, "\x00\x28\x01\x17"); // 40 .. 279 (240 columns)
+        caset[0] = 40;   // 240 cols
+        caset[1] = 279;
+        raset[0] = 53;   // 135 rows
+        raset[1] = 187;
+        madctl = MADCTL::COL_ORDER | MADCTL::SWAP_XY | MADCTL::SCAN_ORDER;
       }
 
       if(width == 135 && height == 240) {
-        command(reg::CASET,     4, "\x00\x34\x00\xba"); // 52 .. 186 (135 rows)
-        command(reg::RASET,     4, "\x00\x28\x01\x17"); // 40 .. 279 (240 columns)
+        caset[0] = 52;   // 135 cols
+        caset[1] = 186;
+        raset[0] = 40;   // 240 rows
+        raset[1] = 279;
+        madctl = 0;
       }
+
+      // Byte swap the 16bit rows/cols values
+      caset[0] = __builtin_bswap16(caset[0]);
+      caset[1] = __builtin_bswap16(caset[1]);
+      raset[0] = __builtin_bswap16(raset[0]);
+      raset[1] = __builtin_bswap16(raset[1]);
+
+      command(reg::CASET,     4, (char *)caset);
+      command(reg::RASET,     4, (char *)raset);
+      command(reg::MADCTL,    1, (char *)&madctl);
     }
 
     // the dma transfer works but without vsync it's not that useful as you could
@@ -160,7 +168,7 @@ namespace pimoroni {
   }
 
   void ST7789::flip(){
-    madctl[0] ^= 0xC0;
-    command(reg::MADCTL,    1, madctl);
+    madctl ^= MADCTL::ROW_ORDER | MADCTL::COL_ORDER;
+    command(reg::MADCTL,    1, (char *)&madctl);
   }
 }
 ```
 
Note also that I've done the same with `CASET` and `RASET`, storing them as `uint16_t` because that is- after all- what the values are. These are then byteswapped and converted to `char *` when sent. It takes me hours to wrap my head around this stuff every time I do it... but the payoff in code clarity is huge.

Also my rotation settings don't quite work- I think some rotations require a memory offset, just for fun.

**Comment by Gadgetoid** [comment]
Above changes are in this commit which you can grab as a patch - https://github.com/pimoroni/pimoroni-pico/commit/55bd4dd02fc67a702811da1ac88021f5b69713e9?branch=55bd4dd02fc67a702811da1ac88021f5b69713e9&diff=unified

If you want to play with it, you can apply it like so:

```
wget https://github.com/pimoroni/pimoroni-pico/commit/55bd4dd02fc67a702811da1ac88021f5b69713e9.patch
git am 55bd4dd02fc67a702811da1ac88021f5b69713e9.patch
```

Just need to figure out all the possible rotation combinations of these registers now :grimacing:

Gadgetoid:
Switching to #103 to finish this off (merging #103 should, in theory, flag this PR as merged anyway).

PR #87, opened 2021-03-08
Display thermo example Keep last 48 temperatures

slabua:
Keep history for the last 48 (240px / 5px) temperatures readings instead of completely clearing the screen once full.


Gadgetoid:
Love it! Thank you.

PR #88, opened 2021-03-08
Add ulab to included modules

Gadgetoid:
This change submodules in our fork of ulab with an added usermod.cmake and adds it to the top-level usermod.cmake for our shipped modules.

It should - in theory - result in ulab being included in release builds.


callimero:
Wow, great. I did some speed tests using the second core rendering mandelbrot sets and then tried (unsuccsessfull) to include ulab. So that will open Micropython to more demanding calculation tasks.

PR #93, opened 2021-03-17
Driver sgp30 Air Quality sensor

simon3270:
In a break from previous drivers, I wrote this one from scratch using the Sensirion data sheet. I believe that it captures the useful functions (and the less-than-useful one that measures raw values).

Measurement starts correctly, and readings do seem to vary as you breathe over the sensor! I've compared it with your Raspberry Pi Python library (https://github.com/pimoroni/sgp30-python), and results are consistent (though I did spot a bug in that code - issue raised).

I think I've been consistent in formatting, but please feel free to tidy it if you decide to merge the code. As before, it's based on your breakouts-dev branch.

### DIFF
--- a/drivers/sgp30/README.md
+++ b/drivers/sgp30/README.md
@@ -1,0 +1,105 @@
+# The SGP30 Environmental Sensor
+
+The SGP30 environmental sensor breakout gives CO2 (eCO2) and Volatile Compound (TVOC) readings.
+
+## Code description
+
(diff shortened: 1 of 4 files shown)


Gadgetoid:
Can confirm, readings vary when I breathe. At least I know I'm alive now!

Cracking job with this one, thank you! And exhaustive documentation too.

ZodiusInfuser:
Thanks @simon3270!

PR #98, opened 2021-03-21
Create Balls_Demo.py

dsssssssss9:
This is adapted from the demo.py in the examples/pico_display directory.
It now displays full screen on a Pico Explorer


Gadgetoid:
Thank you!

Issue #99, opened 2021-03-21
update() is crashing the Pico with a Pico Display attached


nickwebcouk:
This is probably the worst issue I've had to raise as I have no idea what's happening.

I have 300-odd lines of hacked together python code that almost works - but every time I call display.update() the whole pico crashes. Theres no update, there's no screen change, there's no output on the console - it just freezes.

I'm certain a variable hasn't been set correctly but I can't see any issues. Theres no errors - the board just crashes. Because its Micropython I Cant figure out how to enable any advanced logging or debugging so I've included my code here.

The actual code is in a created function and receives the data through that function, but is reproducible using the variables defined here.

Sorry for the amount of print statements - but it was me trying to get to the bottom of what's happening. All variables are printing correctly, so I've no idea what's happening.

I've tried on both Thonny and Visual Studio Code. The code is being run as its typed, so not saved as main.py yet.

I have the Pico Display hooked up to the board directly, with 4 pins that are not in use repurposed to connect a Neo GPS Module.

```
import time
import picodisplay as display 

width = 240
height = 135
display_buffer = bytearray(width * height * 2)  # 2-bytes per pixel (RGB565)
display.init(display_buffer)
display.set_backlight(1.0)

counter = 0

previous = 0
duration = 500
led = 0
currentTime = 0


red = 0
green = 0
blue = 255


FdateGPSStr = "21/03/21"
FtimeGPSStr = "22:23:36"
FfixTypeStrA = "No"
FfixTypeStrB = "Warn"
FsatUseInt = 0
FspeedMPHFlt = 0.0


#def updateAttachedDisplay(FdateGPSStr, FtimeGPSStr, FfixTypeStrA, FfixTypeStrB, FsatUseInt, FspeedMPHFlt):  

print(FdateGPSStr + FtimeGPSStr + FfixTypeStrA + FfixTypeStrB + str(FsatUseInt) + str(FspeedMPHFlt))
# Requires the following variables to be passed

# FdateGPSStr     (Passed straight to display)
#   - 21/02/21
# FtimeGPSStr     (Passed straight to display)
#   - 04:58:20
# FfixTypeStrA    Checked and background colour changed on value
#   if No - Red background
#   if 2D - Yellow background
#   if 3D - Green background
# FfixTypeStrB    Checked and background colour changed on value
#   if Warn - Red background
#   if Valid - Green background
# FsatUse         (Passed straight to display)
# TODO:
[...]

PR #111, opened 2021-03-30
Fix for MicroPython hard lock when passing in non-string object to text function

### LINKED ISSUE #58: pico_display
This example code locks up my the pico device when c ~ 35 to 40. If I rem to the display.text line it works fine.
```python
import machine
import picodisplay as display
import time

led_onboard = machine.Pin( 25, machine.Pin.OUT )

width = display.get_width()
height = display.get_height()
display_buffer = bytearray( width * height * 2 )
display.init( display_buffer )
display.set_backlight(1.0)

count = 0
while True:
    led_onboard.toggle()
    
    count += 1
    t = "{:.1f}".format( count / 10 )
    print( t )
    
    display.set_pen( 0, 0, 0 )
    display.clear()  
    display.set_pen( 255, 255, 255 )
    display.text( t, 1, 40, 240, 4 )
    display.update()
    time.sleep( 1.0 ) 
```

ZodiusInfuser:
Now raises an error if a non-string type is passed into the `.text` function to resolve #58 and #99.

This seemed best, rather that attempting to perform type conversion within the function, which could lead to numerous edge cases.

### DIFF
--- a/micropython/modules/breakout_roundlcd/breakout_roundlcd.cpp
+++ b/micropython/modules/breakout_roundlcd/breakout_roundlcd.cpp
@@ -411,22 +411,36 @@ mp_obj_t BreakoutRoundLCD_text(size_t n_args, const mp_obj_t *pos_args, mp_map_t
     
     breakout_roundlcd_BreakoutRoundLCD_obj_t *self = MP_OBJ_TO_PTR2(args[ARG_self].u_obj, breakout_roundlcd_BreakoutRoundLCD_obj_t);
 
-    mp_check_self(mp_obj_is_str_or_bytes(args[ARG_text].u_obj));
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Can confirm, it changes my test from a hard lock to a `TypeError`:

```python
import machine
import picodisplay as display
import time
led_onboard = machine.Pin( 25, machine.Pin.OUT )
width = display.get_width()
height = display.get_height()
display_buffer = bytearray( width * height * 2 )
display.init( display_buffer )
t = 0.0
display.text( t, 1, 40, 240, 4 )
print("Hello World")
```

Produces:

```
>>> %Run -c $EDITOR_CONTENT
Traceback (most recent call last):
  File "<stdin>", line 10, in <module>
TypeError: can't convert 'float' object to str implicitly
```

PR #115, opened 2021-04-04
Fix picokeypad.clear()

DiddyWolf:
Using release 0.1.1 of pimoroni-pico-micropython, the picokeypad.clear() function does not seem to work. Looking through the code, it appears that clear is actually calling init instead. Unfortunately, I'm just starting with micropython and don't have much C experience outside of Arduino, so testing this (or confirming I'm reading the C code for this module correctly) is a bit out of my depth.


Gadgetoid:
I'm pretty sure you're reading it correctly. D'oh!

The broken Linux build here is my fault :grimacing:

PR #121, opened 2021-04-08
Set pixels as image, add show_bitmap_1d()

graeme-winter:
# Preamble
I don't know how you folks feel about PR's - if this is not your game, I will not be offended if I see a close with no comments. 

# Proposed API extensions

## Show images

Show images using a `set_pixels` method:

```python
image = bytearray(0 for j in range(width * height))
picoscroll.set_pixels(image)
```

seems convenient for cases where you want to redraw the image from one "frame" to the next, means all of the pixel value copying is done in C not Python/ 

## Scroll bitmaps

Scroll bitmaps (one dimensional, e.g. text) using a `show_bitmap_1d` method:

```python
bitmap = bytearray(j for j in range 127)
for offset in range(-17, 127):
    picoscroll.show_bitmap_1d(bitmap, 16, offset)
    picoscroll.update()
```

Detailed illustration in [this gist](https://gist.github.com/graeme-winter/ff08123ceae76399791413f2564eecaa)

Motivation for second one: pico scroll did not offer a simple method to actually scroll text across the display, and I feel that this is pretty fundamental so decided to offer one. I wondered about adding the code to render the text to a bitmap to the API but decided on balance to keep things simple, as fonts are something people hold strong opinions about. 

End product:

https://youtu.be/XIvKc523NwM

Will welcome any feedback here, I tried to stick to the "house style"

### DIFF
--- a/micropython/modules/pico_scroll/README.md
+++ b/micropython/modules/pico_scroll/README.md
@@ -10,6 +10,7 @@ We've included helper functions to handle every aspect of drawing to the matrix 
   - [get_width](#get_width)
   - [get_height](#get_height)
   - [set_pixel](#set_pixel)
+  - [set_pixels](#set_pixels)
   - [update](#update)
   - [clear](#clear)
   - [is_pressed](#is_pressed)
@@ -71,6 +72,17 @@ This function sets a pixel at the `x` and `y` coordinates to a brightness level 
 
 ```python
 picoscroll.set_pixel(x, y, l)
+```
+
+### set_pixels
+
+This function sets all pixel at once from a `bytearray` image indexed
+as `y * picoscroll.get_width() + x`, containing brightness levels
+between 0 and 255. Changes will not be visible until `update()` is called.
+
+```python
+image = bytearray(0 for j in range(width * height))
+picoscroll.set_pixels(image)
 ```
 
 ### update
--- a/micropython/modules/pico_scroll/README.md
+++ b/micropython/modules/pico_scroll/README.md
@@ -85,6 +85,24 @@ image = bytearray(0 for j in range(width * height))
 picoscroll.set_pixels(image)
 ```
 
+### show_bitmap_1d
+
+Show a view of a bitmap stored as the 7 least significant bits of
+bytes in a `bytearray`, top-down. Individual pixels are set to
+`brightness` based on individual bit values, with the view defined by
+the offset and the width of the scroll (i.e. 17 columns). Changes will
+not be visible until `update()` is called.
+
+```python
+bitmap = bytearray(j for j in range 127)
+for offset in range(-17, 127):
+    picoscroll.show_bitmap_1d(bitmap, 16, offset)
+    picoscroll.update()
+```
+
+will scroll a binary counter across the display (i.e. show `0x00` to
+`0x7f` in binary).
+
 ### update
 
 Pushes pixel data from the Pico to the Scroll Pack.  Until this function is called any `set_pixel` or `clear` calls won't have any visible effect.
--- a/micropython/modules/pico_scroll/pico_scroll.c
+++ b/micropython/modules/pico_scroll/pico_scroll.c
@@ -20,6 +20,7 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_0(picoscroll_get_width_obj, picoscroll_get_width)
 STATIC MP_DEFINE_CONST_FUN_OBJ_0(picoscroll_get_height_obj, picoscroll_get_height);
 STATIC MP_DEFINE_CONST_FUN_OBJ_0(picoscroll_update_obj, picoscroll_update);
 STATIC MP_DEFINE_CONST_FUN_OBJ_3(picoscroll_set_pixel_obj, picoscroll_set_pixel);
+STATIC MP_DEFINE_CONST_FUN_OBJ_1(picoscroll_set_pixels_obj, picoscroll_set_pixels);
 STATIC MP_DEFINE_CONST_FUN_OBJ_0(picoscroll_clear_obj, picoscroll_clear);
 STATIC MP_DEFINE_CONST_FUN_OBJ_1(picoscroll_is_pressed_obj, picoscroll_is_pressed);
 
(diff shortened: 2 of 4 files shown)


graeme-winter:
Ugh, I seem to have mixed tabs and spaces, sorry. 

If you have some cleanup I should run, happy to do this (I did not see a developer guide) - am used to running `black` on Python code or `clang-format` on C/C++ -> don't usually pay too much attention to the type of white space.

graeme-winter:
🤔

Having just displayed 

```
x128 = bytearray(range(128))
for offset in range(-17, len(hello)):
    scroll.show_bitmap_1d(hello, 16, offset)
    scroll.update()
    time.sleep(0.1)
```

I think I have this upside down, so my font map is upside down 🙄

Guess I should have done that first -> will revise.

graeme-winter:
Turns out this is probably largely redundant having looked at 

https://github.com/pimoroni/scroll-phat-hd

which would probably be a better API if ported (though still does not allow for a scrolling view across a larger buffer). Does contain a useful 5x7 font though (found in HackSpace issue 41)

Gadgetoid:
PR's - especially ones this thorough and well considered - are always welcome. Thank you! I'll have to give this a try, but I like what you're doing here. Your premise is solid - with "scroll" in the product name but not demonstrated in the software things feel a bit silly.

It could be well worth baking the 5x7 font into the C++ library and adding functions for rendering text across an arbitrary length buffer. There are definitely differences of opinion when it comes to fonts, but I guess having one baked-in wouldn't prevent anyone from ignoring it and rolling their own.

Given the relatively small memory overhead, maybe we could add a "scroll_text" function that just takes a text string, a speed and a loop count and just does the rest for you. It may be possible to render the text even on the fly and avoid a full-sized buffer altogether.

graeme-winter:
@Gadgetoid  thanks for the positive feedback - am working on adding in a C implementation of the font rendering I used in the Python code, so should see a commit added to the PR soon. Adding a `scroll_text` function based on that would be pretty straightforward... 

Along these lines, I guess something for showing text (e.g. 14°C or whatever) in a non-scrolling manner would also be useful. May as well get all this done in one go.

graeme-winter:
I have added `show_text()` now, some doc strings to go with this. This is the full character set:

fontmap

I welcome any further suggestions. 

I would like to run the whole thing through e.g. `clang-format` to clean up the indentation / white space etc. but holding off in case you have tools for this.

graeme-winter:
@Gadgetoid I have added `scroll_text` with this API:

```
text = "Hello, world!"
brightness = 8
delay_ms = 100
scroll.scroll_text(text, brightness, delay_ms)
```

It's not set to loop, as I figured you could do that externally if you wanted. It will scroll the text for an offset of `-17` to `6 x len(text)` i.e. coming from right, towards left, invisible to invisible. I also have a `show_text()` method added which is designed to show "one frame" though as you suggest could be done with on-the-fly rendering. If I get a chance, I will refactor that, as it's currently a little inefficient if you use `show_text()` in a loop as I am rendering bytes which will never be seen. 

However, this is probably in a good place for review now.

graeme-winter:
For info, `show_text()` now only renders what will be shown so the static buffer is one letter + spaces bigger than the view.

Gadgetoid:
> I would like to run the whole thing through e.g. `clang-format` to clean up the indentation / white space etc. but holding off in case you have tools for this.

I wanted to make this part of the GitHub workflow so we could enforce some consistency on the whole codebase... I know line endings have gone all awry at least. An initial tidy would probably be part of the commits adding that change and probably worth doing separately.

Thanks for all the effort! I'll have a play with this now.

Gadgetoid:
D'oh I cannot for the life of me find a Pico Scroll Pack so I've put one on order and will get to this ASAP!

graeme-winter:
> D'oh I cannot for the life of me find a Pico Scroll Pack so I've put one on order and will get to this ASAP!

🤣

Gadgetoid:
Okay having got this up and running, it works great but it should be implemented in the C++ libraries so both C++ Pico SDK and MicroPython users can use it.

I've done that for `scroll_text` in this commit - https://github.com/Gadgetoid/pimoroni-pico/commit/f85115125590aaf129d592e581f041e025c74c75

I've stubbed out `picoscroll_show_text` pending the same treatment.

This, in theory, should have no appreciable difference to Python users but makes this good stuff available to everyone else too.

Gadgetoid:
Okay that commit is stale now, I've force pushed to https://github.com/Gadgetoid/pimoroni-pico/commit/b3798cd7711b9305000407c0a17357665a21edd6 which isn't quite so broken!

Gadgetoid:
Aaand again. I might have got carried away here - https://github.com/Gadgetoid/pimoroni-pico/commit/3861267e8b844bf605192835ec51cce487c9d0cc

Crux of this is making everything use the same underlying utility function - `set_bitmap_1d` so there's no unnecessary code repetition,

Brings show_text (now set_text), show_bitmap_1d (now set_bitmap_1d) and set_pixels into the fold, too.

graeme-winter:
I'll checkout your version a bit later... happy to see merged with C library

graeme-winter:
Picked, built & run - seems to do the same things 🙂

Having in C++ API makes more sense but was not something I was in a position to test. 

How do you want to proceed with this now? Am happy to push the commit to this PR?

Gadgetoid:
Sure, you can push it to this PR- any way works for me.

I had the scrolling text running on my desk for a day or so, rock solid. I'll have to see if it's pin compatible with :eyes: pico wireless pack :eyes:.

I merged the Matrix11x7 driver today and couldn't help but think all this code should probably be ported over. Will have to add that to The List.  :laughing:

graeme-winter:
> Sure, you can push it to this PR- any way works for me.
> 
> I had the scrolling text running on my desk for a day or so, rock solid. I'll have to see if it's pin compatible with 👀 pico wireless pack 👀.
> 
> I merged the Matrix11x7 driver today and couldn't help but think all this code should probably be ported over. Will have to add that to The List. 😆

It's pushed - happy to see this be recycled as much as anyone likes! I only have the scroll top though

graeme-winter:
> Woohoo! Merged. Will get this out in a Python release ASAP!

\o/

PR #122, opened 2021-04-11
Update MicroPython build step and reorder it to after toolchain step.

WayneKeenan:
RaspberryPi have updated their build instructions, e.g. the git repo is no longer valid.


Gadgetoid:
Ooof,  yeah that repository died *some time* ago, and I think I was too busy trying to recover the USER_C_MODULES stuff to notice the wider implications. Thank you!

Issue #124, opened 2021-04-14
 1.3" SPI Colour Round LCD (240x240) Breakout does not work on Rpi pico 


metanav:
I have compiled and flashed the breakout_roundlcd demo from the examples (C++) to Rpi Pico successfully. I have followed the pin map from the drivers/st7789/st7789.hpp. After flashing and resetting the display does not show anything. When I press reset button the display shows some colorful still graphics momentarily. I believe connection is OK since it displays something on keep pressing RESET button (Captain Resetti).

metanav:
With the same pin mappings as stated above it works by adding
` display.set_backlight(225);` after display.init();
I guess PWM gets activated only after set_backlight() call.

Gadgetoid:
Haha! I wish I'd seen this 35 minutes ago because I just had this revelation after *much* head scratching and updated #128 

I have added `set_backlight` to the example *and* into the library as a default value to avoid this surprise in future!

Thank you.

PR #128, opened 2021-04-19
Add breakout-garden SPI slot selection for ST7789 displays

Gadgetoid:
A new constructor has been added which accepts a enum type BG_SPI_SLOT which can be either:

`ST7789::BG_SPI_FRONT` or `ST7789::BG_SPI_BACK`.

This selects the correct CS/BL pins for the respective breakout-garden SPI slot.

Note: the pin designations do not appear to match the BG schematic, but they work!?

This is not a fix for #124 but was for the complexities I observed while looking into it. Took me half an hour to get the darn LCD working even knowing I'd done it before!

Probably needs adding for regular square, too? Not sure how that's all structured.

### DIFF
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -43,6 +43,7 @@ namespace pimoroni {
       pwm_set_wrap(pwm_gpio_to_slice_num(bl), 65535);
       pwm_init(pwm_gpio_to_slice_num(bl), &cfg, true);
       gpio_set_function(bl, GPIO_FUNC_PWM);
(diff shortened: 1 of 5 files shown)


Gadgetoid:
After some considerable head scratching, followed by some head-desking the weird backlight pin issue transpired to be simple:

Correctly setting up the backlight pin meant that the PWM would turn the display backlight *off* and the example was not seeing any backlight value, so it would stay off. Switching to the *other* pin would "fix" the problem by breaking it better. I have corrected the pins back and set a backlight value by default to avoid future surprises.

PR #129, opened 2021-04-20
Add a common header for pins and settings

Gadgetoid:
* Add a common/pimoroni.hpp to list default pins for various add-ons
* Move the BG SPI Slot enum here for safe keeping
* Switch all GPIO pin references to "uint" to match Pico SDK and bring back PIN_UNUSED as UINT_MAX


Gadgetoid:
Oh I just realised I have the AS7262 on my "didn't work" pile, so this isn't quite merge ready yet.

Gadgetoid:
Turns out trying to init the I2C address as the interrupt pin might cause some issues :laughing:

PR #132, opened 2021-04-29
Fixes colours in the MP cheerlights example

helgibbons:


Gadgetoid:
Ooof, how did I even....

Thank you!

PR #134, opened 2021-05-04
C++ and MP support and example for Trackball breakout 

ZodiusInfuser:
Only one example for this. Do we want any more?


Gadgetoid:
Wow you dropped this one out of the blue...

It's... kinda tricky to do examples for I guess, but perhaps something that uses Pico Explorer's display might be cool.

I guess the other obvious one is to make it work over USB as an actual input device... totally academic and practically useless, but cool if we can make it work.

ZodiusInfuser:
Ha, this one was pretty much done last month, just lost in the depths of the big branch. An example for the PicoExplorer's not a bad shout, maybe with a menu of colours you can scroll over and select the trackball to be. I'll give it a thought.

I'd be inclined to avoid anything with USB HID in C at this point, given the many hoops a friend had to go through to get a simple 2 switch keyboard working.

Gadgetoid:
> I'd be inclined to avoid anything with USB HID in C at this point, given the many hoops a friend had to go through to get a simple 2 switch keyboard working.

I'll take your word for it, I was getting dangerously close to giving it a try but I need another timesink like I need a headache.

Otherwise this looks good to me- builds fine, Python and C examples both run. Happy to merge now and we can follow up with examples in good time if anyone has some grand ideas.

ZodiusInfuser:
> I'll take your word for it, I was getting dangerously close to giving it a try but I need another timesink like I need a headache.

Here's my friends repo if you do every want to get your head around it (could be useful for Keybow) but not a priority now: https://github.com/guruthree/pico-superkey-board

> Otherwise this looks good to me- builds fine, Python and C examples both run. Happy to merge now and we can follow up with examples in good time if anyone has some grand ideas.

Cool. Give me till the end of the day to try a PicoExplorer example before merging. I'm currently binding the LTR559.

ZodiusInfuser:
Better C++ example of trackball now added. Tested on PicoExplorer, but should work on DisplayPack too.

Gadgetoid:
That's a niiiiiiceeeeeee demo! You're :fire: today :laughing:

PR #135, opened 2021-05-04
C++ and MP support and example for LTR559 breakout 

ZodiusInfuser:
Only one example at the moment. May do a PicoExplorer example later today whilst doing trackball.

### DIFF
--- a/drivers/CMakeLists.txt
+++ b/drivers/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(esp32spi)
 add_subdirectory(ltp305)
+add_subdirectory(ltr559)
 add_subdirectory(st7789)
 add_subdirectory(msa301)
 add_subdirectory(rv3028)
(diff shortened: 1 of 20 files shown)


ZodiusInfuser:
Okay. A PicoExplorer example isn't getting done today, so maybe just merge this for now?

Gadgetoid:
I'm seeing a much lower proximity output when using this sensor with our C++ driver versus our pure Python one... I'm not sure what accounts for this difference. In both cases the readings are pretty sensible.

Otherwise, everything seems good!

ZodiusInfuser:
Are you meaning the update rate is lower? Because I did add a 100ms time delay to the demos to prevent serial output getting overloaded. Could probably do 50 or 20ms.

ZodiusInfuser:
Spotted some minor issues with my micropython example so pushed an update. For some reason my git lost upstream, so introduced that extra "merge branch" commit.

PR #136, opened 2021-05-05
C++ and MP support and example for SGP30 breakout

ZodiusInfuser:
I breathed on it and values went up, so guess that means it works??


Gadgetoid:
*waves wand of git rebase +2*

wait, wtf, why did that break the tests? aaaaaaaaaaaaaaaaaa

Gadgetoid:
Takes some time to warm up, but eventually starts responding to my breathy attempts to affect the reading.

Baseline level seems to be rock solid at 400, which I guess is PPM equivalent for the baseline CO2 levels in air... I'm practically outdoors in this office so it'd seem logical.

PR #137, opened 2021-05-05
C++ and MP support and example for 1.3" LCD breakout

ZodiusInfuser:
Creating this as a draft @Gadgetoid, as there's very noticable flickering of the LCD that I don't understand. This occurs both in C++ and in Micropython.

Also there's the question of if we are happy with the name `breakout_colourlcd240x240` or whether we'd prefer something shorter such as `breakout_squarelcd`. This would similarly apply to the 0.96" LCD, which is called `breakout_colourlcd160x80` but could be renamed `breakout_rectlcd`. Thoughts?


Gadgetoid:
I think we'll have to stick with the long names- we never know when another display will drop that breaks a shorter naming scheme. Nothing wrong with being precise, even if it's a bit wordy.

I'll see if I can see the flicker- I don't recall having an issue in Python/Pi so it might be an SPI baudrate thing or a backlight thing.

Gadgetoid:
Riiiiight I think I've got this in a sensible place now. I've rebased upon main so it's merge ready, but also added a few commits:

1. Fix for ST7789 display flicker - reduce the SPI baud rate from 60MHz down to 16MHz. The maths - assuming my frazzled brain got it even slightly correct - says this is good for around 17fps.
2. Extra init, taken from the Python library - doesn't seem to make a difference, but have added for completeness
3. Reduced the number of circles in the demo, since it was getting frame drops and looking a bit meh

Let me know if it works for you!

ZodiusInfuser:
> Let me know if it works for you!
Cool. Will try it out on Monday

ZodiusInfuser:
Just tested and flickering is much reduced, so this is good to commit.

Shame about having to lower the number of shapes, as I would have thought this display would have been identical to our 1.5" screen. Anyway, something for another day/month.

PR #154, opened 2021-05-14
Cleanup i2c pins and deinit i2c on init failure

Gadgetoid:
MicroPython doesn't offer any way for C modules to clean up after themselves and often requires a hard reset to get out of a bad state.

This patch attempts to deinit i2c and pins when a device init fails, this allows a user to fail to init a device (with the wrong pins)
and continue to try the correct initialisation without first having to *hard* reset their board.


Gadgetoid:
Individual libraries should *not* be managing I2C state, and not only due to the issues this causes with MicroPython.

This change is shelved in favour of a class to manage the I2C bus across libraries, added to this PR: https://github.com/pimoroni/pimoroni-pico/pull/129

PR #160, opened 2021-05-24
Rewrite this code for #158

### LINKED ISSUE #158: Who wrote the sample code??? (rgb_keypad)
Lately I bought a pi pico and I tried the sample code for the rgb_keypad (Python + C++)

Take this snipped from https://github.com/pimoroni/pimoroni-pico/blob/main/examples/pico_rgb_keypad/demo.cpp#L36-L47
```c++
  uint8_t button = 0;
  for(uint8_t find = 0; find < pico_keypad.NUM_PADS; find++) {
    // check if this button is pressed and no other buttons are pressed
    if(button_states & 0x01) {
      if(!(button_states & (~0x01))) {
        lit |= 1 << button;
      }
      break;
    }
    button_states >>= 1;
    button++;
  }
```
A for-loop with "find" which is never used but a variable "button" that is incremented inside of for - WHY???
Pure documentation, spaghetti code and a lot of shift operations - OMG. Please write better samples and document your code.

Gadgetoid:
Rewrites the cursed Pico RGB Keypad code so that it can't hurt anyone else.

Turns out the whole loop could have been a bitwise operator.

The true lesson was the people we met along the way.

Fixes #158

### DIFF
--- a/examples/pico_rgb_keypad/demo.cpp
+++ b/examples/pico_rgb_keypad/demo.cpp
@@ -5,71 +5,70 @@
 
 #include "pico_rgb_keypad.hpp"
 
+/*
+  Test buttons/lights by lighting them up in a sequence of colors.
+
+  Press each button in turn, it will light up green.
+  Once all buttons are lit, they will turn off.
+  Press them again to light them in the next color.
+*/
+
 using namespace pimoroni;
 
 PicoRGBKeypad pico_keypad;
 
-int main() {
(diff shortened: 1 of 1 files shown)


bennuttall:
Who wrote this code???

Gadgetoid:
> Who wrote this code???

WJDK!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #161, opened 2021-05-25
MP Fix for text scaling on lcd breakouts not working, and non-string vars hard-locking

ZodiusInfuser:
Text code was copied from Display and Explorer, so didn't account for the extra `self` var. Also, Square and Rect LCDs lacked the fix from the Round LCD for non-string vars hardlocking.

### DIFF
--- a/micropython/modules/breakout_colourlcd160x80/breakout_colourlcd160x80.cpp
+++ b/micropython/modules/breakout_colourlcd160x80/breakout_colourlcd160x80.cpp
@@ -398,7 +398,7 @@ mp_obj_t BreakoutColourLCD160x80_character(size_t n_args, const mp_obj_t *pos_ar
     int y = args[ARG_y].u_int;
 
     Point p(x, y);
-    if(n_args == 4) {
+    if(n_args == (ARG_scale + 1)) { // Assumes scale is the last argument
         int scale = args[ARG_scale].u_int;
         self->breakout->character((char)c, p, scale);
     }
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Small remaining, bug - I think you need to count keyword args when detecting the use of "scale" because I can use:

```python
display.text("Hello World", 0, 0, False, scale=3)
```

which will run, but wont scale, versus:


```python
display.text("Hello World", 0, 0, False, 3)
```

which *does* scale.

Also using "MP_QSTR_wr" as an argument name:

https://github.com/pimoroni/pimoroni-pico/blob/dbbe71702a080b4e2cd5388989bef7faaea69d56/micropython/modules/breakout_colourlcd160x80/breakout_colourlcd160x80.cpp#L418

Results in a slightly confusing error:

```python
Traceback (most recent call last):
  File "<stdin>", line 61, in <module>
TypeError: 'wr' argument required
```

These could probably all stand to have the same positional/keyword argument duality as the old i2c constructors, plus `wrap` could quite happily have a default value.

ZodiusInfuser:
Urg, I will have another look at this tomorrow then unless you're already on it?

`Also using "MP_QSTR_wr" as an argument name:` Btw, in case you didn't guess, the var's called `wr` rather than `wrap` because wrap is a reserved word like `start_server` was

Gadgetoid:
> wrap is a reserved word like `start_server` was

:facepalm: 

I'd be tempted to go with longform "wordwrap" if that's acceptable. I think "wr" is maybe a little oblique! :laughing:

ZodiusInfuser:
Sure, that sounds reasonable

PR #164, opened 2021-06-05
Fix character positioning for 3 breakout LCD's in Micropython

frankdrn:
x position was being set by y variable so characters only printed in a diagonal line.


Gadgetoid:
> x position was being set by y variable so characters only printed in a diagonal line.

I'm torn between :facepalm: and :laughing: 

Thank you for PR'ing a fix to this! Looks like I'm going to have to tag a release today!

PR #179, opened 2021-07-20
Add RGBLED and Button libraries

Gadgetoid:
"RGBLED" PWMs 3 pins as a single RGB LED and exposes methods to set the colour via HSV/RGB.

"Button" handles tracking the state and changed state of a single GPIO button, in addition to supporting auto-repeat for held-down buttons.


ZodiusInfuser:
Thought, would these be better to move into drivers rather than libraries?

ZodiusInfuser:
Btw, this is missing Polarity from pimoroni_common.hpp

Gadgetoid:
I think everything is in place now, the downside of just adding these libraries is that there's no example to consume them and actually throw errors if they don't compile.

I have rebased #181 from this, though, so there's a good chance it's in working order!

PR #180, opened 2021-07-20
Fix some ST7789 regs

Daft-Freak:
Fixes frame rate/porch register confusion and uses `PWCTRL1` instead of 0xd0. Probably the only thing from my pile of hacks that counts as a fix :smile:

(Also, the `PORCTRL` and `FRCTRL2` values look like the defaults... Maybe a few of the others too, didn't check.)


Gadgetoid:
Thank you! I'd actually noticed the weird `PORCTRL` naming when bringing up another LCD.

PR #222, opened 2021-11-21
Fix documentation examples setting RGB LED

### LINKED ISSUE #221: Micropython Plasma docs reference non-existent function set_led
In the [onboarding docs for the Plasma module](https://github.com/pimoroni/pimoroni-pico/tree/main/micropython/modules/plasma) the function for setting an LED is documented as 

```python
led_strip.set_led(0, 255, 0, 255)
```

Replicating the documentation examples results in 

```
AttributeError: 'plasma_ws2812' object has no attribute 'set_led'
```

The correct function would be `set_rgb`

```python
led_strip.set_led(0, 255, 0, 255)
```

eddhurst:
Update the examples using set_led function and replacing to use set_rgb

Fixes #221


Gadgetoid:
Thank you!

PR #225, opened 2021-11-30
Driver for the mlx90640 thermal camera

Gadgetoid:
I don't know how useful this will be, or how reliable since this camera intensely dislikes me, but I realised the 32x24 sensor is a good fit for a 32x32 or 64x64 matrix panel and it all went downhill from there.

May or may not work with other display products- uses about 8k of RAM.

## TODO
* [ ] Python bindings


Gadgetoid:
Rebased, since I think this could be interesting to use with Galactic or Cosmic.

PR #235, opened 2022-01-04
Fix "repeat_time" docs in README

mozz100:
Fix documentation around `Button` class's `repeat_time` behaviour.


Gadgetoid:
Good catch, thank you!

PR #246, opened 2022-02-05
I2C Library for boilerplate projects?

lordmortis:
So I was trying [the boilerplate project](https://github.com/pimoroni/pico-boilerplate) with the [plasma2040 rotary encoder example](https://github.com/pimoroni/pimoroni-pico/blob/main/examples/plasma2040/plasma2040_rotary.cpp) and it looked like it was trying to link with a library called `pimoroni-i2c` 

There wasn't one, so I made this cmake file to create one, and then could compile the example boilerplate. 

You can check out my boilerplate project [here](https://github.com/lordmortis/pimoroni-plasma-test) - maybe I was setting it up wrong? 

I think the linker is trying to link that library because of [this line](https://github.com/pimoroni/pimoroni-pico/blob/91b3ffb7d81d94009b5ca16ccaf9f0f23d808041/drivers/ioexpander/ioexpander.cmake#L10) in `ioexpander.cmake` ?

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,9 +1,1 @@
-add_library(pimoroni_i2c INTERFACE)
-
-target_sources(pimoroni_i2c INTERFACE
-  ${CMAKE_CURRENT_LIST_DIR}/pimoroni_i2c.cpp)
-
-target_include_directories(pimoroni_i2c INTERFACE ${CMAKE_CURRENT_LIST_DIR})
-
-# Pull in pico libraries that we need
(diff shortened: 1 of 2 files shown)


lordmortis:
@Gadgetoid so something more like what i've pushed now?

Gadgetoid:
@lordmortis oh wow, you're quick on the update. Exactly that, yes. Thank you!

lordmortis:
You got me right as work finished and it was a pretty trivial change. Yay timezones!

PR #259, opened 2022-03-02
Adding support for Servos

ZodiusInfuser:
This is still a work in progress, but enough of it is functional to make its presence known for those who may want to have a play. Features several user facing C++/MP classes:

- **Servo**: Hardware PWM based servo implementation
- **ServoCluster**: 30-channel PIO based servo implementation
- **Calibration**: An object containing calibration data for a servo. This is common to both implementations.

ServoCluster todo:
- [x] Add MP constructor that accepts a list/tuple of pins
- [x] Auto-configure phase offsets (to reduce peak current draw)
- [X] Support multiple instances, and generally do PIO management better
- [x] Ensure objects clean up nicely
- [x] Make underlying PWMCluster class support 100% duty cycle

Other todo:
- [x] Documentation
- [x] MP Examples
- [x] C++ Examples

Future features:
- [ ] Delay frequency change until next cycle
- [ ] Be able to set an interrupt pin
~~- [ ] Support for servo easing~~ (Added a servo easing example)


ZodiusInfuser:
Making PWMCluster support 100% PWM was easier than I thought (but required much testing), so I did it first even though it was last on the list 😆

ZodiusInfuser:
PWMCluster now supports offsets that cause Pulses to go beyond the wrap level. This required a bunch of extra logic to deal with edge cases as pulse lengths change, but I think I got them all 🤞 .

This was a prerequisite for implementing `Auto-configure phase offsets (to reduce peak current draw)`, as now it will be possible to evenly space all servos across their time window, even when running at frequencies above 50Hz (again 🤞 )

ZodiusInfuser:
Can now create multiple instances of ServoCluster, and classes now get deleted cleanly-ish. There can be some weirdness in both these cases that I currently do not have enough understanding of DMA bringup/teardown to be able to diagnose.

ZodiusInfuser:
Well, this is as ready as it will ever be 😅

Gadgetoid:
LGTM!

*deep breath*

PR #260, opened 2022-03-02
Badger2040 C++ library and example fixes

MichaelBell:
Hi,

I'm not sure if you take pull requests, but I've made a couple of fixes to the Badger2040 library to make the C++ examples work correctly.

Main fix is that because `badger.update()` doesn't block, before halting the example should spin waiting for `badger.is_busy()` to return false before powering down, otherwise the last update effectively doesn't happen when running off battery.

The other fixes I've made are to:
- Set pull down on button E, this didn't obviously seem to be causing me a problem but it was odd that it was missing.
- Set the wake button states, as used by the image example (but I don't block waiting for the button to be released)
- Change `update_button_states()` to set the button states to the current value, rather than ORing in pressed buttons to the existing value, so that it is possible to check if a button is currently pressed without using `wait_for_press()`
- Implement the `button_states()` function so it is possible to test the masked states as a user.

Note I haven't tested if these fixes cause any problems in python land (if this library is even used by python - I haven't looked at that at all).

Cheers, Mike

### DIFF
--- a/examples/badger2040/badger2040_image.cpp
+++ b/examples/badger2040/badger2040_image.cpp
@@ -56,6 +56,9 @@ int main() {
   }
 
   badger.update();
+  while (badger.is_busy()) {
+    sleep_ms(10);
+  }
   badger.halt();
 
 }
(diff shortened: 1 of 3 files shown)


MichaelBell:
No worries, I understand C++ is not the primary target for the Badger.  Initially I thought that the battery wasn't providing enough power to allow the display to update - but once I worked out what was going on I was happy to make fixes!

Looking again I see the `update()` method actually has a parameter to choose to be blocking, so using that would be a better fix for the examples.

PR #275, opened 2022-03-06
badger2040: Set default clock to a valid date

lurch:
2022 isn't a leap year, so there is no 29th February!


Gadgetoid:
JUSTICE FOR LEAP YEAR BABIES

PR #276, opened 2022-03-06
badger2040: Only 24 hours in a day (not 60)

lurch:
...at least on my planet :earth_africa:  :laughing:


Gadgetoid:
In my defense I still think the whole concept of time is a scam to sell expensive watches and I've never quite got to grips with it.

lurch:
Unfortunately the RP2040 is a little bit too small to make an effective sundial :slightly_frowning_face:

PR #278, opened 2022-03-06
badger2040: prevent date being set to an invalid value

lurch:
(by restricting the maximum number of days in a month as necessary)

The RP2040 datasheet says "The RTC does not check that the programmed values are in range. Illegal values may cause unexpected behaviour."

### DIFF
--- a/micropython/examples/badger2040/clock.py
+++ b/micropython/examples/badger2040/clock.py
@@ -22,6 +22,12 @@ button_b = machine.Pin(badger2040.BUTTON_B, machine.Pin.IN, machine.Pin.PULL_DOW
 button_c = machine.Pin(badger2040.BUTTON_C, machine.Pin.IN, machine.Pin.PULL_DOWN)
 
 
+def days_in_month(month, year):
+    if month == 2 and ((year % 4 == 0 and year % 100 != 0) or year % 400 == 0):
+        return 29
+    return (31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31)[month - 1]
+
+
 # Button handling function
 def button(pin):
     global last, set_clock, cursor, year, month, day, hour, minute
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you. I was *really* hoping I could be lazy and that something somewhere in the MicroPython API or SDK would magically take care of this. Gosh I hate times & dates. Note the glaring lack of a calendar example.

micropython/examples/badger2040/clock.py
```diff
@@ -22,6 +22,11 @@
 button_c = machine.Pin(badger2040.BUTTON_C, machine.Pin.IN, machine.Pin.PULL_DOWN)
 
 
+def days_in_month(month, year):
+    if month == 2 and ((year % 4 == 0 and year % 100 != 0) or year % 400 == 0):
+        return 29
+    return (31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31)[month - 1]
+
```
Gadgetoid, line 29: Linting says no:

> micropython/examples/badger2040/clock.py:31:1: E302 expected 2 blank lines, found 1

lurch:
>  I was really hoping I could be lazy and that something somewhere in the MicroPython API or SDK would magically take care of this.

With all the "embedded" stuff being so minimal and "light weight", it looks like the MicroPython layer was assuming that the hardware-layer handles it, and the RP2040 hardware assumes that some higher-level handles it? :man_shrugging:

micropython/examples/badger2040/clock.py
lurch: Hurpfully fixed...

lurch:
> Gosh I hate times & dates.

https://en.wikipedia.org/wiki/Swatch_Internet_Time :wink:

PR #285, opened 2022-03-09
Make USB powered and battery powered Badger2040 work the same

MichaelBell:
One annoyance with the badger is difference in the behaviour of `halt()` between running with power connected and without.

I realised this would be simple to fix - instead of just looping forever in halt(), instead wait for a button press and then reset - this should give you exactly the same behaviour as when running from battery.

I've also included a couple of other small changes:
- A fix to make convert.py handle images that are already black and white
- And I hope you might indulge me and allow a change from the internal state of the Badger2040 class from private to protected, which would help me continue to use your class as a base for my LowPowerBadger.

### DIFF
--- a/examples/badger2040/image_converter/convert.py
+++ b/examples/badger2040/image_converter/convert.py
@@ -24,8 +24,11 @@ options = parser.parse_args()
 def convert_image(img):
     if options.resize:
         img = img.resize((296, 128)) # resize
-    enhancer = ImageEnhance.Contrast(img)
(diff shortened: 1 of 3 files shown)


Gadgetoid:
I'm going to be honest- I *barely* touched the C++ stuff outside of ripping it to pieces in order to make it work for MicroPython. T'was all very last minute, seat of my pants. As such I'm really grateful for your efforts here- thank you!

I will gladly indulge you!

PR #287, opened 2022-03-10
Error handling for data_to_py import

### LINKED ISSUE #283: Badger 2040 - Fix instructions for using convert.py in Thonny
The most recent version of `convert.py` includes a call to `import data_to_py` which causes an error if you're trying to run `convert.py` from Thonny as per the [instructions here](https://learn.pimoroni.com/article/getting-started-with-badger-2040#customising-the-badgeros-examples)

You can comment out `import data_to_py` to get it to work.

helgibbons:
Lets you run `convert.py` without `data_to_py` being present.

Fixes #283

### DIFF
--- a/examples/badger2040/image_converter/convert.py
+++ b/examples/badger2040/image_converter/convert.py
@@ -11,7 +11,7 @@ from PIL import Image, ImageEnhance
 from pathlib import Path
 try:
     import data_to_py
-except ImportError as error:
+except ImportError:
     pass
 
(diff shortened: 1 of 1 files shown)


helgibbons:
Never mind - yours is nicer @Gadgetoid :D

Gadgetoid:
This is roughly how I started until I realised there are only a few dozen lines to copy over 🤣 figured it would make sense to compact it all.

Issue #272, opened 2022-03-04
16 MB Pico LIPO is limited to 2MB in Micro Python?


alphanumeric007:
I'm using a 16 MB Pimoroni Pico LIPO with the custom Pimoroni uf2 file and its limiting me to the default Pi PICO 2 MB. It appears there is a need for a custom Pimoroni Pico Lipo Micro Python uf2 file, to get access to the full 16 MB?

[ https://forums.pimoroni.com/t/16-mb-pico-lipo-only-showing-4mb/19018/13 ]

helgibbons:
Just adding my datapoints to this one in case they're helpful! 

There's been some debate at Pirate Towers about whether the MicroPython file system expands automatically to fill the available flash memory. This doesn't seem to be happening for me (or if it is, some other shenanigans are afoot!) - I'm not able to upload a file that's larger than 1.3MB to my 4MB Pico LiPo, and `os.statvfs` seems to be reporting a total filesystem size of 1.375MB on both a Pico LiPo and a standard Pico.

Error when uploading a big file with Thonny:
![Screenshot 2022-03-07 150021](https://user-images.githubusercontent.com/50950368/157649667-1bbc09b4-9358-4048-921e-aae80bec36a1.jpg)

`os.statvfs` results:
![Screenshot 2022-03-07 162741](https://user-images.githubusercontent.com/50950368/157649718-4cae9912-04aa-4deb-b0fd-81f4bdd72bcb.jpg)
.

alphanumeric007:
I posted on the Pi Foundation Forum asking about making use of more than 2MB and was directed here.    https://micropython.org/download/?port=rp2   Listed there are custom images for the 8MB Tiny 2040 , Pico Lipo 4Mb and Pico Lipo 16MB.  I flashed the Tiny 2040 uf2 to it and was able to upload a 5.1MB mp3 file without error. The % copied actually worked too.  =) I’m guessing these images don’t have any of the files Pimoroni adds to their images? 
https://micropython.org/download/PIMORONI_TINY2040/
https://micropython.org/download/PIMORONI_PICOLIPO_4MB/
https://micropython.org/download/PIMORONI_PICOLIPO_16MB/

Gadgetoid:
> I’m guessing these images don’t have any of the files Pimoroni adds to their images?

They don't. I wonder if they could. Would save us having to build them all again.

[...]

PR #291, opened 2022-03-11
Build MicroPython for multiple boards

Gadgetoid:
Upgrades the CI to output builds for:

* Pico (2MB)
* Tiny 2040 (8MB)
* Pico LiPo (4MB)
* Pico LiPo (16MB)

Provides the 16MB Flash build requested in #272

I've added some significant caching both via ccache (for caching compiler output) and GitHub Actions cache.

MicroPython and the Pico SDK are cloned, prepped and cached per version- this cache should probably be used by other workflows to speed up these stages since the MicroPython version changes very rarely.

Installing ARM GCC is still a huge time sink- it needs to be done for *each* port from the list above and can take anywhere between 1 and 2 minutes, even hanging for much, much longer- this quickly multiplies up across all builds. Could we grab the zipped toolchain and cache it?

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -14,8 +14,19 @@ jobs:
     runs-on: ubuntu-20.04
     name: Dependencies
     steps:
+    - name: Workspace Cache
+      id: cache
+      uses: actions/cache@v2
+      with:
+        path: ${{runner.workspace}}
+        key: workspace-micropython-${{github.ref}}-${{github.sha}}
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Managed to slash the ARM GCC install time by wget'ing the toolchain, extracting it and adding it to PATH. This is surprisingly easy to do as a drop-in replacement to grabbing it from apt and seems to take an approximately constant 30sec with no drawn-out hangs.

PR #294, opened 2022-03-12
add sleep to help and info badger 2040

japharl:


japharl:
add sleep to badger 2040 about & help commands so people can see the text before screen refereshes to the launcher menu

Gadgetoid:
Whoops, I'd broken these by fixing something else- previously the launcher just crashed when a script exit, so the contents would stay on screen 😬 

Thank you!

Gadgetoid:
Opened as #298 to merge

PR #297, opened 2022-03-14
Extending qrgen.py to allow multiple QR Code files

jpwsutton:
This is a remix on `qrgen.py`, pulling in features from `image.py` that would allow for multiple qr code files that could be cycled through. I don't know if you'd rather keep the `qrgen.py` example "clean", but I'm using this personally on my badger2040 so thought others might like to as well.

On first start, it creates a `qrcodes` directory, then adds the example file in there instead, the acts just like the image example where it draws dots on the right hand side and allows you to cycle through with the up and down buttons.

Signed-off-by: James Sutton <1068763+jpwsutton@users.noreply.github.com>


jpwsutton:
I've pulled from upstream/main and have integrated the battery improvement changes. This also fixes what I think is a small mistake in the current qrgen.py where the LED is turned on permanently (https://github.com/pimoroni/pimoroni-pico/blob/522c83dc196883eb13b0a6b944711f8efe5f3892/micropython/examples/badger2040/qrgen.py#L30)

I've included the LED as it's nice to have, but only turn it on when the display is updating.

Gadgetoid:
The LED *should* turn off as soon as Badger 2040 hits the "halt" method, indicating that power is cut.

Under any other circumstances staying lit is the preferred behavior, since it shows that the system is still running and using power- even if it might not be doing much. This will happen on USB power- be it from a host computer or USB battery- since in this case Badger loses its ability to cut its own power.

I think something has gone wrong with your rebase here, too, since GitHub's indicating there are 34 changed files and 53 commits!

You might need to create a fresh branch from main, manually patch over your changes and force push it up.

Otherwise, your changes look pretty sensible! Thank you. Albeit I have not had a moment to test them yet.

(Actually keeping the LED lit is *such* a good litmus test of whether a given app is using power or not, I'm almost tempted to steal it entirely from the user and force it always on. Or - at least- never allow the user to set it below a specific brightness.)

(we should have included a power LED...)

PR #310, opened 2022-03-22
Added an example of a memory game.

waterphox:
This is the second game I tried making for the badger2040.
It's a simple simon-style memory game where you must memorize the order that shapes appear on the screen.
When the game says GO!,  just press the corresponding buttons to eliminate the shapes!


Gadgetoid:
Awesome. Thank you. On my TODO list to give this a try!

If you'd like this to be a builtin example, the "eliminationsquares.bin" should be added to git as the source image, and CMake will take care of building it if you add a line like the following: https://github.com/pimoroni/pimoroni-pico/blob/fa2774b1147257e537e91c29f4ee270459d2b2e4/micropython/examples/badger2040/micropython-builtins.cmake#L38

Source image should probably be included either way! And you should use the existing "assets" directory: https://github.com/pimoroni/pimoroni-pico/tree/main/micropython/examples/badger2040/assets

Gadgetoid:
Intrigued to know how you managed folders! But I think I have a pretty good idea how they could work, based on our new stateful apps approach.

Let me get https://github.com/pimoroni/pimoroni-pico/pull/313 merged, and a new release sorted, and we can start to improve things from there!

App icons are a tricky business, we might have to make it easy to just store individual image files alongside the app, use a default icon, or something else.

waterphox:
Basically just some folder names, and bumping out the examples list in to sub-lists.
`currentMenu = 0
previousMenu = 0
menuTitle = ("MiniGLaDOS v2", "Folders", "Crap", "Utilities", "Games")

selections = [
    #Main Menu
    [
        ("_badge", 5),
        ("_image", 3),
        ("_fursona", 8),
        ("_list", 4),
        ("_ebook", 2),
        ("_folders", 9)
    ],
    #Folders Menu
    [
        ("_back", 10),
        ("_games", 9),
        ("_utilities", 9),
        ("_crap", 9)
    ],
    #Crap Menu
    [
        ("_back", 10),
        ("_info", 6),....ETC`
And then making the example launching function check for folder names
`def launch_example(index):
    global currentMenu, previousMenu, page
    try:
        pick = selections[currentMenu][(page * 3) + index][0]
        if pick == "_folders":
            currentMenu = 1
            page = 0
            render()
        elif pick == "_crap":
......
elif pick == "_back":
            currentMenu = previousMenu
            previousMenu = previousMenu - 1
            page = 0
            if previousMenu < 0:
                previousMenu = 0
            render()
        else: #They picked something other than a folder or back option
            launch(pick)
        return True
    except IndexError: #Typically triggers if you press a button under a blank spot (no icon/app there)
        return False`

micropython/examples/badger2040/elimem.py
```diff
@@ -0,0 +1,438 @@
...
+            d.icon(squares, 3, 144, 24, columns[2], rows[1])
+        d.partial_update(columns[2], 72, 24, 24)
+    elif sq == 8:  # downtri2
+        activesquares[3][1] = 1
+        if activesquares[2][1] == 1:
+            d.icon(squares, 2, 144, 24, columns[2], rows[1])
+        else:
+            d.icon(squares, 4, 144, 24, columns[2], rows[1])
+        d.partial_update(columns[2], 72, 24, 24)
+    pass
```
lurch, line 198: You seem to have lots of `pass` calls in various places, which seem to be unnecessary?

waterphox: Yes! I am new to Python and was using those to visualize where my functions/sections ended. As it results in no operation, I figured that was OK- though the interpreter does not ignore them like a comment. At this point I can remove them!

Gadgetoid:
Badger has a new home over at https://github.com/pimoroni/badger2040 and this could make a cool default example if we port it!

Issue #311, opened 2022-03-23
CMake file in /libraries/pico_wireless contains an outdated path


deodre:
I was trying to build a custom project for Pico Wireless Pack using the pico_wireless library and I encountered this error:

```
PIMORONI_PICO_PATH is /home/pi/pimoroni-pico
PICO_SDK_PATH is /home/pi/pico/pico-sdk
PICO platform is rp2040.
PICO target board is pico.
Using board configuration from /home/pi/pico/pico-sdk/src/boards/include/boards/pico.h
TinyUSB available at /home/pi/pico/pico-sdk/lib/tinyusb/src/portable/raspberrypi/rp2040; enabling build support for USB.
Using PICO_EXAMPLES_PATH from environment ('/home/pi/pico/pico-examples')
CMake Error at /home/pi/pimoroni-pico/libraries/pico_wireless/pico_wireless.cmake:1 (include):
  include could not find load file:

    /home/pi/pimoroni-pico/libraries/pico_wireless/../../drivers/esp32wireless/esp32wireless.cmake
Call Stack (most recent call first):
  CMakeLists.txt:26 (include)


-- Configuring incomplete, errors occurred!
See also "/home/pi/projects/pico-boilerplate/build/CMakeFiles/CMakeOutput.log".
See also "/home/pi/projects/pico-boilerplate/build/CMakeFiles/CMakeError.log".
```
However this didn't happen when I built the examples from pimoroni. The file responsible for the error is */libraries/pico_wireless/pico_wireless.cmake*:

```
include(${CMAKE_CURRENT_LIST_DIR}/../../drivers/esp32wireless/esp32wireless.cmake)
add_library(pico_wireless INTERFACE)

target_sources(pico_wireless INTERFACE
  ${CMAKE_CURRENT_LIST_DIR}/pico_wireless.cpp
)

target_include_directories(pico_wireless INTERFACE ${CMAKE_CURRENT_LIST_DIR})

# Pull in pico libraries that we need
target_link_libraries(pico_wireless INTERFACE pico_stdlib hardware_spi esp32spi)
```
On the first line, the path that is included is named differently
![image](https://user-images.githubusercontent.com/45459027/159664532-c989e9f1-d7a9-4f94-b5fd-6859825c5d82.png)

After modifying the path in the *pico_wireless.cmake* file the project build correctly.

PR #312, opened 2022-03-23
Updated path to esp32 drivers in /libraries/pico_wireless/pico_wireless.cmake

tcervinski-csie:
I updated the path to the esp32 drivers, responsible for the error specified in Issue #311


Gadgetoid:
Thank you!

PR #314, opened 2022-03-23
Improves the image viewer demo by adding rollover, screen refresh controls, and invert function

waterphox:
The viewer will now jump to image 0 if down is pressed on the last photo, or jump to the last if up is pressed on image 0.
To invert the display, press B.
To switch between TURBO and FAST refresh speeds, press C.
To jump to / switch between MEDIUM and SLOW speeds, press C while holding the USER button.

### DIFF
--- a/micropython/examples/badger2040/image.py
+++ b/micropython/examples/badger2040/image.py
@@ -5,7 +5,6 @@ import machine
 import badger2040
 from badger2040 import WIDTH, HEIGHT
 
-
 README = """
 Images must be 296x128 pixel with 1bit colour depth.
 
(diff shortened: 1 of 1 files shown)


waterphox:
I know that the built-in demos are mainly supposed to be simple(ish) examples of the various functions and functionality of the badge, so you don't want them to be _too_ complex- but I think at least adding the rollover would make the image demo much more satisfying to use.

Gadgetoid:
Thank you! I'm all for extra features as long as they don't over-complicate things, but this is going to conflict with a pretty huge rewrite of parts of  badgerOS so bear with me!

I'm working on today's release, which will change a bunch of things (for the better, generally) and we can go from there if you're willing!

waterphox:
Absolutely! Let me know when the merges are/release is done and I will re-implement the changes in the new images.py and resubmit.

waterphox:
Alright. I'll wait for v1.18.6 and go from there.
The new appstate stuff adds a level of complexity (and coolness), but I'll make sure the inverted state and refresh speed states are stored through it. Though all four speeds will probably just cycle with one button, now that USER is harder to use!
I probably won't update my own badge past 1.18.4. I am either doing something interactive or putting something on the screen and turning it off, so the loss of a button combo isn't worth it!

Gadgetoid:
Okay to move the goalposts again, Badger 2040 and Badger 2040 W are now combined and moved here: https://github.com/pimoroni/badger2040

If you're willing, these changes should be made over there!

PR #316, opened 2022-03-26
Even better Badger wake handling

MichaelBell:
I love the smell of assembler in the morning!  (Well, it was morning when I started).

This change reads the GPIO state and sets the enable 3v3 as early as possible during the boot process.  It shaves a couple of milliseconds off when the buttons are read and the power is enabled, because it's done before the external crystal oscillator is started and stable.

Because the initialization is now done in assembler, I've removed the WakeUpInit class the cpp file, so now there are just trivial functions to read the saved state.

I've also removed the access to pressed_to_wake from the Badger2040 object, as it isn't something apps should be using, and removed clear_pressed_to_wake again as it is unnecessary given the launcher waits for all buttons to be released anyway.

### DIFF
--- a/micropython/examples/badger2040/launcher.py
+++ b/micropython/examples/badger2040/launcher.py
@@ -17,7 +17,6 @@ woken_by_button = badger2040.woken_by_button()  # Must be done before we clear_p
 if badger2040.pressed_to_wake(badger2040.BUTTON_A) and badger2040.pressed_to_wake(badger2040.BUTTON_C):
     # Pressing A and C together at start quits app
     exited_to_launcher = badger_os.state_clear_running()
-    badger2040.clear_pressed_to_wake()
 else:
     # Otherwise restore previously running app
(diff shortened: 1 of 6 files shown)


Gadgetoid:
Wow! And here was me thinking the existing solution was already lightyears ahead of what I could contrive. I wish I'd known about these methods to poke stuff into early boot/init stages.

Gadgetoid:
Works great. Thank you.

It occurs to me that this functionality probably wants copying over to the C++ library at some point, so that C++ project users get some slightly better functionality in this regard. The original intent was to use RP2040s sleep modes, but given this works *so well*, starts up quickly even when bringing up a whole MicroPython VM, and uses (probably considerably) less power than sleeping it makes sense to include it.

MichaelBell:
Yes - we'd want to make this optional for C++ land, as the user might want to do something custom.  But having the same default behaviour switchable with a cmake variable would be nice.

PR #318, opened 2022-03-26
Badger2040: Improve help, info, qrgen

### LINKED ISSUE #317: Badger QR, info and help apps exit on any button press
Apps that just call `halt()` and exit will quit to launcher when any button is pressed.  They should halt in a loop instead.

MichaelBell:
This fixes #317.  I've also tweaked help to try and improve readability and add the a&c exit combo.  

This has lost the subtlety of USER needing to be pressed after the other buttons on battery, but I'm not sure how clear that was anyway.

One thought I had was that new users are likely to go into an app and have no way of knowing how to exit.  Reset no longer gets you out.  So maybe the default launcher state on a fresh Badger should be to launch the help app?

### DIFF
--- a/micropython/examples/badger2040/help.py
+++ b/micropython/examples/badger2040/help.py
@@ -2,7 +2,7 @@ import badger2040
 from badger2040 import WIDTH
 
 TEXT_SIZE = 0.45
-LINE_HEIGHT = 16
+LINE_HEIGHT = 20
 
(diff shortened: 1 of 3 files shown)


Gadgetoid:
I wonder if a first boot overlay in launcher.py with just "press a + b to exit apps" would suffice. Something to think about for sure.

PR #319, opened 2022-03-26
Badger2040: Update list app to use new system

MichaelBell:
Does what it says on the tin!

I added some special handling to clear everything if the list of items was changed at the top of the file, so if a user edits that then they start from scratch.

I also made the update speed slower on page change as a turbo update then leaves the items a bit corrupt.

### DIFF
--- a/micropython/examples/badger2040/list.py
+++ b/micropython/examples/badger2040/list.py
@@ -1,9 +1,11 @@
+import binascii
+
 import badger2040
 import badger_os
 
-# **** Put your list title and contents here *****
+# **** Put your list title here *****
 list_title = "Checklist"
-list_items = ["Badger", "Badger", "Badger", "Badger", "Badger", "Mushroom", "Mushroom", "Snake"]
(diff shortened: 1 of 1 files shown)


MichaelBell:
Currently the docs indicate you can change the list of items by rewriting checklist.txt, which this removes.

Possibly it would be better to always read the list of items from checklist.txt rather than defining them in code.  That would make checklist.txt the user configurable list of items, and independent of the state.  We could do the same thing to notice when the checklist has changed (though maybe it would be more sensible to just store a hash of the item list, rather than the complete list.

MichaelBell:
Modified to use checklist.txt as the list of items, as described above.  It strips off the " X" on the end of item names in user's existing check lists, although it doesn't attempt to preserve the state.

MichaelBell:
Now migrates from old format.

Gadgetoid:
Thank you - I had disappeared down a font rabbit hole!

Will test and release 1.18.6 tomorrow :tada:

PR #324, opened 2022-03-29
Extending qrgen.py to allow multiple QR Code files

jpwsutton:
I made a mess of the merge in PR #297, so have run a clean branch with the new changes on top.

This is a remix on qrgen.py, pulling in features from image.py that would allow for multiple qr code files that could be cycled through. I don't know if you'd rather keep the qrgen.py example "clean", but I'm using this personally on my badger2040 so thought others might like to as well.

On first start, it creates a qrcodes directory, then adds the example file in there instead, the acts just like the image example where it draws dots on the right hand side and allows you to cycle through with the up and down buttons.

Signed-off-by: James Sutton <1068763+jpwsutton@users.noreply.github.com>

### DIFF
--- a/micropython/examples/badger2040/qrgen.py
+++ b/micropython/examples/badger2040/qrgen.py
@@ -1,13 +1,22 @@
 import badger2040
 import qrcode
+import time
+import machine
+import os
+import badger_os
 
-
-# Open the qrcode file
+# Check that the qrcodes directory exists, if not, make it
 try:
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you! And, sorry, a minor change to qrgen has conflicted this. Don't worry- I can sort it if there are no other changes needed.

Gadgetoid:
Rebased, updated to the new API and raised as #333

PR #325, opened 2022-03-29
Badger2040: system_speed call, plus performance improvements

MichaelBell:
I've added a `badger2040.system_speed()` call to allow the user to easily adjust the clock speed to 5 settings, from 3MHz to 250MHz in sensible steps, applying under/over voltage as appropriate.  Lower clock frequencies are not available when USB is connected, but can be used to save power when on battery.

Clearing the screen and drawing large rectangles were relatively inefficient, so I now do these by editing the framebuffer directly.  Rectangles honour the odd behaviour of being affected by the current thickness setting, to preserve backwards compatibility.

### DIFF
--- a/drivers/uc8151/uc8151.cpp
+++ b/drivers/uc8151/uc8151.cpp
@@ -481,15 +481,15 @@ namespace pimoroni {
   uint32_t UC8151::update_time() {
     switch(_update_speed) {
       case 0:
-        return 5500;
+        return 4500;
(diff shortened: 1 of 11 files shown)


MichaelBell:
As discussed, will add Micropython UART reinitialization after changing the clock speed, in case anyone is using hardware UART.

Gadgetoid:
Working beautifully, thank you!

PR #329, opened 2022-03-30
Bitmap Fonts: Port outline font 14.

Gadgetoid:
Make the font data a uint16_t array, this doubles the storage size of existing fonts but allows for bitmap fonts up to 16 pixels tall.

Some kind of variable storage size for each character could be useful, since we need to be careful about storage size. Currently this change adds around 3.7k.

Additionally there's no API support for fixed/variable width fonts yet. That said in most cases a fixed width font with non fixed width letters looks awful anyway...


Gadgetoid:
Managed to squeeze this down to 1.5k by packing the old <= 8 pixel tall fonts nicely into the uint16_t data array, rather than leaving a bunch of empty space.

About 1k of that is the new outline font.

We should probably support setting bytearray or memory view style fonts so MicroPython users can sideload their own.

Gadgetoid:
Replaced by #332

PR #330, opened 2022-03-30
Reinitialize hardware after system clock change

MichaelBell:
Reinitialize UART and SPI after system clock change.

Document that you must change the system clock before initializing any I2C drivers.

### DIFF
--- a/libraries/badger2040/badger2040.cpp
+++ b/libraries/badger2040/badger2040.cpp
@@ -225,6 +225,12 @@ namespace pimoroni {
     }
     w += _thickness - 1;
     h += _thickness - 1;
+    if (x + w > 296) {
+      w = 296 - x;
+    }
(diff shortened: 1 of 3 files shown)


MichaelBell:
Also now fixes a bug introduced in #325 which overran the framebuffer, resulting in Thonny not being able to interrupt Micropython

Gadgetoid:
I've seen a couple of hangups from Thonny where I've needed to press the hardware reset button, but they seem to have got better lately.

This all looks great, though, and it probably goes without saying but bonus points and huge props for updating the documentationl

MichaelBell:
Thanks!

Before a4acf4b, I could never get Thonny to reconnect if I ran the new version of the list app from Thonny  and pressed the up or down arrow once (or more).  With the fix it worked reliably for the duration of me testing my changes, at least!

PR #337, opened 2022-04-07
Update Badger readme to include bitmap fonts

helgibbons:


Gadgetoid:
Good shout, thank you!

PR #341, opened 2022-04-17
Badger 2040 - Edit to badge.py to allow selection from multiple badges in a folder

LittleHobbyShop:
Changed behaviour:

- badges (including default badge.txt) are now stored in the \badges subfolder. Both are created if missing.
- badge images should be named as per text file with "-image.bin" suffix to match the text file.
- State management used. Resets to default badge if state stored badge is removed.
- Similar selection indicator to other menus is added, only smaller as not to interfere with the image.
- Removed reference to Button C which conflicted with the dual press of A+C to return to launcher.

### DIFF
--- a/micropython/examples/badger2040/badge.py
+++ b/micropython/examples/badger2040/badge.py
@@ -1,6 +1,11 @@
 import time
 import badger2040
 import badger_os
+import os
+
+# Multiple badge text and image files supported. All files to be placed in newly created '\badges' folder.
+# Associated image files to be named the same as the text file with '-image' appended.
+# For each text file e.g. 'badge1.txt', an associated image file e.g. 'badge1-image.bin' will load
 
 # Global Constants
 WIDTH = badger2040.WIDTH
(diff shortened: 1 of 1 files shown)


SimonShirley:
@LittleHobbyShop Love the idea of this and thought I'd give it a try, as I'm currently at an event where a second badge would be very useful!!

I'm running into an issue though. When I try to run the script, my badger2040 presents a warning: "Local variable referenced before assignment".

I've managed to pin it down to Line 91:
``` imgfile = str(file).split(".")[0] + "-image.bin" ```

By the point it gets to this code, in my debugging, my badger2040 has 2 badge .txt files, so fails the ``` if ``` condition on Line 86. In doing so, the ``` file ``` variable isn't defined, so when Line 91 is called, the error appears.

I've tested by adding ``` file = BADGES[n] ``` before the call to set ```imgfile``` and in doing so, the badge then appears on the badger2040.

Great work though!

LittleHobbyShop:
@SimonShirley Crikey! I thought this had died! Seeing as it's now 9 months behind I'm not surprised there's a few snags with it. I'll see what I can do soon but it won't be before your event ends I wouldn't think.

Gadgetoid:
@LittleHobbyShop Badger 2040W is posed for launch sometime soon. It has a few breaking changes from Badger 2040, but it should be possible to port your changes over.

I've made the first step of setting up a "badges" directory for the new firmware, but I haven't made the example aware of this since I wanted to merge your PR & credit you first before porting these very worthwhile improvements.

In my case I added a line to the bottom of the ".txt" file to link to a specific image. And in 2040W land (which uses PicoGraphics) it supports JPEG rendering, so it links to a ".jpg" rather than a ".bin" file. No conversion hassle.

The new Badger 2040W badge.py is here - https://github.com/pimoroni/pimoroni-pico/pull/633/files#diff-331fe57c205e2242f8e53d48a6cda824ae80733e85f47d7c2a863b17524b46cf

I hope to update Badger 2040 to match 2040W (switching over to PicoGraphics rather than its own custom setup, and baking in examples as a filesystem rather than hiding them)

Anyway while I totally remembered this PR, I apparently forgot to say- thank you! This is a great idea.

Gadgetoid:
Badger 2040 and Badger 2040 W have been unified and are moving to https://github.com/pimoroni/badger2040

You can find the badge example here - https://github.com/pimoroni/badger2040/blob/main/badger_os/examples/badge.py

(It's the same as the one linked above, but in its new home)

A port of these changes would be appreciated, but if you should find yourself short of time let me know and I'll pull them over.

PR #342, opened 2022-04-18
Added a pico display 2.0 Mandelbrot set example.

martinbudden:
Both pico cores are used: core0 handles the key input and updates the
display at low resolution. Core1 updates the display at high resolution.

For speed, fixed point arithmetic is used. The main loop also checks
for cyles to speed things up.

Six color palettes are provided, including 3 HSV, greyscale and black/white.


Gadgetoid:
Thank you! And sorry for the huuuuge delay getting back about this. It's been quite a... decade?

Rebased, ported to PicoGraphics and raised a replacement PR in #680

PR #344, opened 2022-04-19
fix micropython line and triangle primitives for LCD displays

cedel1:


cedel1:
Fixed (hopefully) all failing linting checks (spaces etc.).

Gadgetoid:
Ouch, some rather shameful oversight in the MicroPython bindings there on our part! Thanks for catching/fixing it.

PR #345, opened 2022-04-23
Updating the samples README to reflect current contents

andypiper:
Added notes on the change to how the QR Code sample works, and that the Conway sample exists. And some other bits and bobs.


Gadgetoid:
Greatly appreciated, thank you!

PR #346, opened 2022-04-24
Fixes/pico graphics

ThomasPDye:
fixes build for external projects using pico_graphics

### DIFF
--- a/libraries/pico_graphics/CMakeLists.txt
+++ b/libraries/pico_graphics/CMakeLists.txt
@@ -1,1 +1,9 @@
-include(pico_graphics.cmake)
+
+add_library(pico_graphics 
+${CMAKE_CURRENT_LIST_DIR}/types.cpp
+${CMAKE_CURRENT_LIST_DIR}/pico_graphics.cpp
+)
+
+target_include_directories(pico_graphics INTERFACE ${CMAKE_CURRENT_LIST_DIR})
+
+target_link_libraries(pico_graphics bitmap_fonts)
--- a/libraries/pico_graphics/pico_graphics.cmake
+++ b/libraries/pico_graphics/pico_graphics.cmake
@@ -1,3 +1,6 @@
+
+include(${CMAKE_CURRENT_LIST_DIR}/../bitmap_fonts/bitmap_fonts.cmake)
+
 add_library(pico_graphics 
     ${CMAKE_CURRENT_LIST_DIR}/types.cpp
     ${CMAKE_CURRENT_LIST_DIR}/pico_graphics.cpp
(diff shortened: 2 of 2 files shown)


Gadgetoid:
I think my preferred solution would be to include "bitmap_fonts/bitmap_fonts.cmake" in the consuming project, but that's because I'm wary of this pattern becoming the norm- since it duplicates the whole library config in two places.

Another possible approach could be a root level include for `pimoroni-pico` which includes *all* libraries and dependencies, since only the ones that are linked into a project will be used anyway. Linking them individually is just busywork, I suppose!

Daft-Freak:
Hmm, I almost opened ~the same PR (but with less duplication):
```diff
diff --git a/libraries/pico_graphics/pico_graphics.cmake b/libraries/pico_graphics/pico_graphics.cmake
index 67fdd79a..07cc65e5 100644
--- a/libraries/pico_graphics/pico_graphics.cmake
+++ b/libraries/pico_graphics/pico_graphics.cmake
@@ -1,4 +1,12 @@
-add_library(pico_graphics 
+if(NOT TARGET bitmap_fonts)
+    include(${CMAKE_CURRENT_LIST_DIR}/../bitmap_fonts/bitmap_fonts.cmake)
+endif()
+
+if(NOT TARGET hershey_fonts)
+    include(${CMAKE_CURRENT_LIST_DIR}/../hershey_fonts/hershey_fonts.cmake)
+endif()
+
+add_library(pico_graphics
     ${CMAKE_CURRENT_LIST_DIR}/types.cpp
     ${CMAKE_CURRENT_LIST_DIR}/pico_graphics.cpp
     ${CMAKE_CURRENT_LIST_DIR}/pico_graphics_pen_1bit.cpp
@@ -13,4 +21,4 @@ add_library(pico_graphics
 
 target_include_directories(pico_graphics INTERFACE ${CMAKE_CURRENT_LIST_DIR})
 
-target_link_libraries(pico_graphics bitmap_fonts hershey_fonts pico_stdlib)
\ No newline at end of file
+target_link_libraries(pico_graphics bitmap_fonts hershey_fonts pico_stdlib)

```

... since I didn't have the best experience trying to do unicorn things: 
- Tried to `include(libraries/galactic_unicorn/galactic_unicorn)` -> compile error due to missing pico_graphics
- Include that -> linker errors due to missing *_fonts
- recreate dependency tree in my code and hope it doesn't change

Having a top-level "all the libraries" include would also work, maybe even have the import file include that. The extra include step per-library seems a bit weird compared to the pico-sdk...

Gadgetoid:
@Daft-Freak I guess that would be sort-of our equivalent of Pico SDK's `pico_sdk_init()` and it makes sense.

Gadgetoid:
Replaced by #686

Thank you both!

PR #365, opened 2022-05-20
MicroPython: Use placement new to alloc classes on GC_HEAP

Gadgetoid:
This change should hopefully allow us to write more generic classes that do not need special treatment under MicroPython.

I have updated Servo/PWM to remove the dynamic buffers in place of constant-sized buffers defined in the class. Using `m_new` to allocate some memory in the GC_HEAP and `placement new` to construct a class within that memory, we can push the *entire* class into MicroPython's memory region and avoid running out of memory.

I have also updated Pimoroni I2C to make this change both in the `machine.I2C` to `PimoroniI2C` promotion and in the regular I2C object.

### DIFF
--- a/drivers/motor/motor_cluster.cpp
+++ b/drivers/motor/motor_cluster.cpp
@@ -9,22 +9,22 @@
 namespace motor {
   MotorCluster::MotorCluster(PIO pio, uint sm, uint pin_base, uint pin_pair_count, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
-                             bool auto_phase, PWMCluster::Sequence *seq_buffer, PWMCluster::TransitionData *dat_buffer)
-    : pwms(pio, sm, pin_base, (pin_pair_count * 2), seq_buffer, dat_buffer, false), pwm_frequency(freq) {
+                             bool auto_phase)
+    : pwms(pio, sm, pin_base, (pin_pair_count * 2), false), pwm_frequency(freq) {
     create_motor_states(direction, speed_scale, zeropoint, deadzone, mode, auto_phase);
   }
 
   MotorCluster::MotorCluster(PIO pio, uint sm, const pin_pair *pin_pairs, uint32_t length, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
-                             bool auto_phase, PWMCluster::Sequence *seq_buffer, PWMCluster::TransitionData *dat_buffer)
-    : pwms(pio, sm, pin_pairs, length, seq_buffer, dat_buffer, false), pwm_frequency(freq) {
+                             bool auto_phase)
+    : pwms(pio, sm, pin_pairs, length, false), pwm_frequency(freq) {
     create_motor_states(direction, speed_scale, zeropoint, deadzone, mode, auto_phase);
   }
 
   MotorCluster::MotorCluster(PIO pio, uint sm, std::initializer_list<pin_pair> pin_pairs, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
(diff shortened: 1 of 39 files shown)


drivers/motor/motor_cluster.cpp
```diff
@@ -10,21 +10,21 @@ namespace motor {
   MotorCluster::MotorCluster(PIO pio, uint sm, uint pin_base, uint pin_pair_count, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
                              bool auto_phase, PWMCluster::Sequence *seq_buffer, PWMCluster::TransitionData *dat_buffer)
```
ZodiusInfuser, line 12: The seq_buffer and dat_buffer constructors can be removed from here too

drivers/motor/motor_cluster.cpp
```diff
@@ -10,21 +10,21 @@ namespace motor {
...
-    : pwms(pio, sm, pin_base, (pin_pair_count * 2), seq_buffer, dat_buffer, false), pwm_frequency(freq) {
+    : pwms(pio, sm, pin_base, (pin_pair_count * 2), false), pwm_frequency(freq) {
     create_motor_states(direction, speed_scale, zeropoint, deadzone, mode, auto_phase);
   }
 
   MotorCluster::MotorCluster(PIO pio, uint sm, const pin_pair *pin_pairs, uint32_t length, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
                              bool auto_phase, PWMCluster::Sequence *seq_buffer, PWMCluster::TransitionData *dat_buffer)
```
ZodiusInfuser, line 19: And from here

drivers/motor/motor_cluster.cpp
```diff
@@ -10,21 +10,21 @@ namespace motor {
...
-    : pwms(pio, sm, pin_pairs, length, seq_buffer, dat_buffer, false), pwm_frequency(freq) {
+    : pwms(pio, sm, pin_pairs, length, false), pwm_frequency(freq) {
     create_motor_states(direction, speed_scale, zeropoint, deadzone, mode, auto_phase);
   }
 
   MotorCluster::MotorCluster(PIO pio, uint sm, std::initializer_list<pin_pair> pin_pairs, Direction direction,
                              float speed_scale, float zeropoint, float deadzone, float freq, DecayMode mode,
                              bool auto_phase, PWMCluster::Sequence *seq_buffer, PWMCluster::TransitionData *dat_buffer)
```
ZodiusInfuser, line 26: Last one

Gadgetoid:
Changed ALL THE THINGS!

Gadgetoid:
@ZodiusInfuser I have updated this PR to use some new macros:

```c++
#define m_new_class(cls, ...) new(m_new(cls, 1)) cls(__VA_ARGS__)
#define m_del_class(cls, ptr) ptr->~cls();m_del(cls, ptr, 1)
```

Which allow you to just do:

```c++
self->breakout = m_new_class(BreakoutAS7262, (pimoroni::I2C *)(self->i2c->i2c), args[ARG_int].u_int);
m_dell_class(BreakoutAS7262, self->breakout);
```

This avoids having to remember the ugly syntax for placement new and the multiple steps for a correct delete.

Gadgetoid:
See also:

https://github.com/pimoroni/pimoroni-pico/blob/bc0390b86c9879de0298f4cd94f8be21ddd534dc/micropython/modules/util.hpp

Which collects all the various nonsense we were repeating on a module by module basis into a single utility include.

ZodiusInfuser:
Cool! That `MP_OBJ_TO_PTR2` Could probably be renamed to something without the 2 in, since that was my lazy name and we've ended up running with it for over a year 😬

Gadgetoid:
We can change that in the next refactor :laughing: but yes I suppose it makes sense as part of the C++ suite of macros I'm proposing upstream, too.

Main reason for not renaming it- I just can't think of a better name!

PR #370, opened 2022-05-23
Small pico unicorn cleanups

Daft-Freak:
Just some minor namespace pollution and float stuff. This saved me ~1k on my already tiny binary that I didn't need to optimise.

Looks like similar gamma stuff is also used in other places, trying to avoid changing a bunch of other stuff though. (The "c++ way" of fixing this would be `#include <cmath>` and `std::pow`, but went with the smaller change).

Also slightly annoying is the three identical gamma LUTs wasting 1k of RAM... but I don't need that either so :shrug:

### DIFF
--- a/libraries/pico_unicorn/pico_unicorn.cpp
+++ b/libraries/pico_unicorn/pico_unicorn.cpp
@@ -177,7 +177,7 @@ namespace pimoroni {
           uint8_t row_select_mask = ~(1 << (7 - row));
           bitstream[row_select_offset] = row_select_mask;
 
-          uint16_t bcd_ticks = pow(2, frame);
+          uint16_t bcd_ticks = 1 << frame;
           bitstream[bcd_offset + 1] = (bcd_ticks &  0xff00) >> 8;
           bitstream[bcd_offset]     = (bcd_ticks &  0xff);
         }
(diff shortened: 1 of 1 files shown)


Gadgetoid:
1k you say? That's good news for our MicroPython build if it has the same effect. I'll need to test this!

Thank you

Gadgetoid:
Looks like it saves ~1544 bytes on the MicroPython build.

Before:

```
Memory region         Used Size  Region Size  %age Used
           FLASH:      590900 B       640 KB     90.16%
             RAM:      238780 B       256 KB     91.09%
       SCRATCH_X:          0 GB         4 KB      0.00%
       SCRATCH_Y:          0 GB         4 KB      0.00%
   text	   data	    bss	    dec	    hex	filename
 590796	    104	 205292	 796192	  c2620	/home/phil/Development/pico/micropython-v1.18/ports/rp2/build-PICO/firmware.elf
```

After:

```
Memory region         Used Size  Region Size  %age Used
           FLASH:      589356 B       640 KB     89.93%
             RAM:      238748 B       256 KB     91.08%
       SCRATCH_X:          0 GB         4 KB      0.00%
       SCRATCH_Y:          0 GB         4 KB      0.00%
   text	   data	    bss	    dec	    hex	filename
 589252	    104	 205292	 794648	  c2018	/home/phil/Development/pico/micropython-v1.18/ports/rp2/build-PICO/firmware.elf
```

I need to somehow get this magic into GitHub Actions so it can notify build size changes.

Gadgetoid:
Works fine in MicroPython, plus the flash saving. Much appreciated! Thank you.

As for the *three* LUTs and the sometimes-there's-a-gamma-table-include mess... uhmm. Yeah.

Daft-Freak:
Well, that was a nice bonus. (Not sure where it came from, would've thought there were still other uses of `pow`... but okay)

PR #372, opened 2022-05-25
Support for parallel ST7789 displays

Gadgetoid:


Gadgetoid:
Ended up being a duplicate of #373 which builds upon this.

PR #395, opened 2022-06-20
Updated PicoGraphics README and .gitignore

kevinmcaleer:
I've fixed the typo in the rotation example it had the parameter: 

```python
display = PicoGraphics(display=PICO_DISPLAY, roation=90)
```

which I've fixed with:

```python
display = PicoGraphics(display=PICO_DISPLAY, rotate=90)
```
---
I've also added a bit more information about how to use the sprite sheets after fiddling with that for a bit and then figuring it out.
---
Finally I've added two exclusions to .gitignore:
`.DS_Store` which is some file system cruft that macOS insists on creating in each folder, and also `venv` which is the default folder when creating virtual environments `python -m venv venv`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@
 
 **/build
 .vscode
+
+# Apple filesystem cruft
+.DS_Store
+venv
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Congrats on your (first?) PR! Thank you.

kevinmcaleer:
Yes - my very first PR! Thanks @Gadgetoid, for merging so quickly.

PR #398, opened 2022-06-21
MICS6814: read_nh3 and read_oxidising now call the correct get_raw_* functions

ahnlak:
As per DrDozzy's comments on Discord, the MICS6814 breakout code uses 'get_raw_red' for both nh3 *and* oxidising readings too, rather than the correct ones.

Note: I don't actually have one of these breakouts, so my testing has been limited to 'does it still compile' :smile:


Gadgetoid:
It's a testament to how completely opaque these gas values are that a mistake this blatant can survive for over a year :roll_eyes:  thank you for PR'ing a fix!

PR #411, opened 2022-06-27
added midpt-ellipse

Toastee0:
Draws elipses. midptellipse(int x radius, int y radius, Point)
Re-try, after clean rebase.

### DIFF
--- a/libraries/pico_graphics/pico_graphics.cpp
+++ b/libraries/pico_graphics/pico_graphics.cpp
@@ -116,6 +116,113 @@ namespace pimoroni {
       }
     }
   }
+  
+void PicoGraphics::midptellipse(int rx, int ry, Point &p)
+                  
+{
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I think this will explode since all surface manipulation must now go via `set_pixel` for a single pixel or `set_pixel_span` for a horizontal span. The actual framebuffer is now `void *` and isn't guaranteed to be in any particular format or layout.

Toastee0:
I'll rewrite it using set pixel, and a check bounds function.

Gadgetoid:
This has been languishing for a while, any interest in bringing it kicking and screaming into 2023 PicoGraphics?

Gadgetoid:
Note to self that adding Ellipse to #783 might be a good idea.

Closing this for now, feel free to re-open if you decide to tackle it!

PR #422, opened 2022-07-02
Pico Graphics: Double buffer scanline conversion

MichaelBell:
This fixes an issue where the Tufty display was mildly corrupted when using Pico Graphics modes other than RGB565.

The problem was the DMA transfer in ST7789 would read from the same buffer that the scanline conversion was setting up for the next line.  This meant that the first few pixels of each row would have the values from the next row, and could possibly cause further corruption if the RP2040 was otherwise busy (e.g. when USB is enabled).

### DIFF
--- a/drivers/st7735/st7735.cpp
+++ b/drivers/st7735/st7735.cpp
@@ -172,8 +172,10 @@ namespace pimoroni {
       gpio_put(dc, 1); // data mode
       gpio_put(cs, 0);
 
-      graphics->scanline_convert(PicoGraphics::PEN_RGB565, [this](void *data, size_t length) {
-        spi_write_blocking(spi, (const uint8_t*)data, length);
+      graphics->frame_convert(PicoGraphics::PEN_RGB565, [this](void *data, size_t length) {
+        if (length > 0) {
+          spi_write_blocking(spi, (const uint8_t*)data, length);
+        }
       });
 
       gpio_put(cs, 1);
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -290,8 +290,10 @@ namespace pimoroni {
       spi_write_blocking(spi, &cmd, 1);
       gpio_put(dc, 1); // data mode
 
-      graphics->scanline_convert(PicoGraphics::PEN_RGB565, [this](void *data, size_t length) {
-        spi_write_blocking(spi, (const uint8_t*)data, length);
+      graphics->frame_convert(PicoGraphics::PEN_RGB565, [this](void *data, size_t length) {
+        if (length > 0) {
+          spi_write_blocking(spi, (const uint8_t*)data, length);
+        }
       });
   
       gpio_put(cs, 1);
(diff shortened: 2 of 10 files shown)


Gadgetoid:
Sorry for the lack of response here, Pico W has been causing me all kinds of headaches.

I'm apprehensive about allocating too much RAM here since I keep running into issues with stack/heap usage outside of MicroPython's gc_heap just quietly blowing up and hardlocking MicroPython.

It may be we need to refactor the scanline conversion callback to (optionally?) accept a buffer that can be allocated on gc_heap with `m_new`. In most drawing contexts this buffer will also be constant, so it probably makes sense to allocate it up front and avoid any chance of memory fragmentation.

I've been running builds with `-Wstack-usage=` to try and figure out where we're introducing potential pitfalls, and it warns on all the `scanline_convert` function with:

```
warning: stack usage might be unbounded [-Wstack-usage=]
```

Which given the dynamic nature of the buffer, is understandable!

Sorry to avalanche you with this information in a really helpful PR. I don't expect you to move heaven and earth to make these changes, but I need somewhere to braindump so I can remember this!

MichaelBell:
Hi!  Yes I saw that tweet and issue where you needed to move a large stack allocation on to the heap for Pico W compatibility.  Is the problem the wifi interrupt handler using lots of stack?  That seems like it could cause everyone issues.

Anyway - yes this change is doubling the stack usage from 640 bytes to 1280, which is quite a lot.  I agree the easiest fix would be to just alloc these buffers on the heap along with the frame buffer when a non-565 Pen is constructed.

A more complex fix would be to chase the DMA in the scanline conversion, which would avoid increasing the memory usage at all.

MichaelBell:
Hi!  Yes I saw that tweet and issue where you needed to move a large stack allocation on to the heap for Pico W compatibility.  Is the problem the wifi interrupt handler using lots of stack?  That seems like it could cause everyone issues.

Anyway - yes this change is doubling the stack usage from 640 bytes to 1280, which is quite a lot (though I think the stack is 4KB by default so it's not totally unreasonable).  I think the easiest fix would be to just alloc these buffers on the heap along with the frame buffer when a non-565 Pen is constructed.

But maybe a better fix would be to chase the DMA in the scanline conversion, which would avoid increasing the memory usage. I did a quick test and seem to have that working - if you're happy that we can use the existing 640 bytes of stack here, I can tidy it up.

Gadgetoid:
Chasing the DMA would be preferable and is what I was going for. I'd hope to factor out the 640 bytes at some point and allow MicroPython to allocate and pass it in, but that's a future problem :laughing: 

We also want to refactor the 16-bit RGB565 pen type to *not* be byteswapped, so the internal memory representation uses native endianess and... not horribly confusing. This will require all existing 16-bit displays to be scanline converted even in RGB565 mode to do the native to device byteswap.

MichaelBell:
I went for a simpler approach of just transferring 64 pixels at a time, so this will always use 256 bytes of stack regardless of your display width.  Note that I have only tested this on the Tufty2040 so I haven't tested the SPI transfer mode.

Because this function now converts a whole frame, 64 pixels at a time, I renamed it `frame_convert` given it has nothing to do with scanlines any more.  That unfortunately meant the previous hack for detecting when the frame was finished (to determine whether the transfer should block) didn't work any more, so I added a new hack ;)

I also refactored the conversion to use a common method on a base class to avoid repeating the code in 3 locations.  This is probably a bit slower, but the data transfer to the display is still going to dominate.

Gadgetoid:
Sheesh has it been nearly a month already!?

This changeset fixed a flicker in our factory tests, but I need to test across some other products and make sure there are no obvious regressions.

Using DMA always makes me *slightly* nervous, but it's largely only used for displays anyway... and there are plenty of DMA channels :laughing: 

Will aim to merge today. Whew. And thank you!

Gadgetoid:
Looks like in PEN_P4 mode there's a slight left-to-right wrap happening on Pico Display. I'll have to dig and see why-

image

Above you can see the very left-most pixels of the "H" wrap over to the right.

Gadgetoid:
Fixed the Pico Display bug (stemming from an underlying bug in Pen P4 as noticed by @MichaelBell) with https://github.com/pimoroni/pimoroni-pico/commit/bb9b5b69d3287e3a645e3e11804b7e425b28bd4e

And made a slight improvement to the RGB565 frame convert (At least I think it's an improvement) in https://github.com/pimoroni/pimoroni-pico/commit/fc3060b640bc1218d0cdf9bc65c719b9c449b121

Gadgetoid:
Very happy with this, thank you! The RP2040 RAM gods smile upon us :laughing:

MichaelBell:
Great to get this fix in!   Agree those changes look good - thank you.

PR #433, opened 2022-07-08
Create show_ip_address

lesley-byte:
To show the ip address on Inky pack when creating the LED is ON/ LED is OFF server for the pico W.  Showing the IP address on the inkypack may make it more useful. To more people.  This solves the big problem of turning it into a server as well as shows that you can change the inky pack screen and not just turn the onboard led on and off.


Gadgetoid:
I like it- thanks! Unpacks some of the more complex stuff I hid away and serves as a good, in-depth example. Nice work.

PR #442, opened 2022-07-12
Add Cheerlights example for Pico W / Pico Explorer

helgibbons:
Maybe there's a better place to put `network_manager.py` so we don't have to copy it to every examples directory?


Gadgetoid:
I hope to put `network_manager.py` into a pypi package, since I keep subtly changing it.

PR #447, opened 2022-07-15
Autow web example

rabid-inventor:
Web UI example ready for merge


Gadgetoid:
Thank you, I've pushed this onto https://github.com/pimoroni/pimoroni-pico/pull/445 for merging.

PR #458, opened 2022-07-22
CI: Revert: wget ARM GCC toolchain

Gadgetoid:
This change resulted in an incomplete toolchain producing broken, bloated builds with exceptions and introspection enabled.


Gadgetoid:
Of particular note, `d_print_comp_inner` was making its way into the build. I'm guessing the lack of the ARM stdlib probably caused havoc here. Pulling the whole arm GCC toolchain seems to work, but it's *huge* and takes just as long as grabbing from apt. I guess hosted runners might be the only way to speed this up?

PR #473, opened 2022-07-27
Badger2040: Produce build without baked-in modules.

Gadgetoid:


datakurre:
@Gadgetoid Any chance that this somehow produced both Badger2040 release builds without the baked-in modules? See: https://github.com/pimoroni/pimoroni-pico/releases/tag/v1.19.6

MarcKarasek:
So how does this work?  Will badger just look for main.py in the top level dir?

Gadgetoid:
Yep. That's how MicroPython works in general, anything you save as "main.py" is run at boot.

PR #480, opened 2022-08-01
MicroPython: Early wakeup GPIO latch module.

Gadgetoid:


Gadgetoid:
Looks like `make BOARD=PICO_W -C ports/rp2 submodules` is not working here and I can't figure out why. Have reverted to our own manual submodule checkout which pulls down a load of stuff we don't necessarily want, but since it's cached anyway... not really a huge deal.

PR #485, opened 2022-08-04
BadgerOS: Fix with-modules build.

### LINKED ISSUE #484: Badger 2040 got stuck with the latest firmware v1.19.6
I upgraded to `pimoroni-badger2040-v1.19.6-micropython.uf2` and it acts like the OS is not there — I can program the screen and LED, but the OS never comes up. Downgrading to `v1.19.5` seems to have solved the issue. Could it be that it’s the wrong file in the firmware, the one without the Badger OS?

Gadgetoid:
Should fix #484

### DIFF
--- a/.github/workflows/micropython-badger2040.yml
+++ b/.github/workflows/micropython-badger2040.yml
@@ -100,19 +100,19 @@ jobs:
       shell: bash
       working-directory: micropython/ports/rp2
       run: |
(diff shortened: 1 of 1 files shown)


sctv:
The artifact works for me (badger2040 with os)

Gadgetoid:
Thank you!

PR #487, opened 2022-08-04
ST7789 Parallel: Work with SYS clock faster than default

MichaelBell:
When I tried to overclock my Tufty the display went odd (it actually split in two taking up the top half of the screen - seems the display hardware copes OK with the clock being two fast - it just misses every other byte, or something like that).

This PR makes PIO setup take account of the current system clock when specifying the divider.  Using a maximum of 32MHz is a fairly conservative guess at where the safe limit is - my Tufty works OK at 40MHz.  The old default was 31.25MHz if you kept a 125MHz system clock.

### DIFF
--- a/drivers/st7789/st7789.hpp
+++ b/drivers/st7789/st7789.hpp
@@ -5,6 +5,7 @@
 #include "hardware/gpio.h"
 #include "hardware/pio.h"
 #include "hardware/pwm.h"
+#include "hardware/clocks.h"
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you. A default overclock is something I'm really keen to add (upstream support to MicroPython at least), specially for our first-party boards like Tufty where we're sticking an RP2040 onboard. Worked out pretty fine for PicoSystem

PR #491, opened 2022-08-09
Add adjust_to_sea_pressure function

adriangalilea:
Now describe_pressure() gives meaningful insights if you set the altitude

Verified it works here:
https://keisan.casio.com/exec/system/1224575267

Verified that it matches the pressure of my iPhone weather app for my location.

If altitude is not set it returns the same value so the altitude variable doesn't need to be edited and it works just as before.


Gadgetoid:
Thank you!

PR #493, opened 2022-08-11
MicroPython: Ported PicoSleep module.

Gadgetoid:
This is a very experimental port of @ghubcoder's "picosleep" - https://github.com/ghubcoder/micropython-pico-deepsleep/

I have vendored Pico Extras `sleep.c` and `sleep.h` because life's too short to deal with extraneous submodules or repository clones for a couple of source files that are going to be version pinned anyway.


Gadgetoid:
Welp this was an unmitigated failure plagued by weird build issues that happen on CI and not my local machine... and the resulting build doesn't boot!? I don't even.

*steps quietly away*

PR #514, opened 2022-09-16
set pen for text to 15 for better visibility

dreambucket13:
I am using the inky pack. When setting the pen to 1 to write text against a black background (pen = 0), the text is not legible.  Setting it to 15 (white) is much more legible and serves as a much better demo.


Gadgetoid:
Good catch, thank you!

This didn't get updated when I changed PicoGraphics to fancy dithering.

dreambucket13:
Nonl problem! I really like the inky display.  Next step getting it going
with freeRTOS.

PR #520, opened 2022-09-26
Correct common path

gauntface:
I noticed this link doesn't work and I believe this should fix it.


helgibbons:
Super, thanks - those relative links are tricksy!

Gadgetoid:
Thank you!

PR #525, opened 2022-10-02
Move set_brightness to APA102

PeterCopeland:
Documentation for set_brightness function was under WS2812 but text referred to APA102, and set_brightness is not supported on WS2812. I've moved the paragraph to the end of the APA102, verbatim.


Gadgetoid:
Good catch! Thank you.

PR #531, opened 2022-10-05
Adds digital clock example for pico_display

hoganri:
Starts at 12:00 and allows the user to adjust the hours/minutes using the pico_display buttons

### DIFF
--- a/micropython/examples/pico_display/pico_clock.py
+++ b/micropython/examples/pico_display/pico_clock.py
@@ -1,0 +1,85 @@
+import time
+import qrcode
+from pimoroni import Button
+from picographics import PicoGraphics, DISPLAY_PICO_DISPLAY, PEN_P4
+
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you!

You should keep `currentTimeHours` and `currentTimeMinutes` as integers, and use Python's string formatting to display and right-justify them. EG:

```python
>>> "{:02}:{:02}".format(9, 30)
'09:30'
```

That would save a lot of unnecessary conversion to/from int.

There are also a few linting errors, which our CI will tell you about here - https://github.com/pimoroni/pimoroni-pico/actions/runs/3193502590/jobs/5218789053

There's also a clock example for Badger 2040 which you might want to draw inspiration from - https://github.com/pimoroni/pimoroni-pico/blob/main/micropython/examples/badger2040/clock.py

hoganri:
Thanks for the feedback. I believe each of these items are addressed in the new commit.

PR #542, opened 2022-10-21
Auto2040w mp pwm

rabid-inventor:


Gadgetoid:
Looks like this is conflicted to all heck, does it need rebased or did it get merged via another branch?

ZodiusInfuser:
I think it's because this PR predates Auto 2040 W Mini, and so all the examples / docs have differing descriptions.

PR #546, opened 2022-10-25
CheerLights example for the Galactic Unicorn

Corteil:
Simple IoT (internet of things) example code.

CheerLights MicroPython client.

You will need the secrets file with your network details, in the same folder as cheerlights.py to run on start rename cheerlights.py to main.py


Gadgetoid:
Thank you!

Couple of changes -

Our readme generally says to use `WIFI_CONFIG.py` from the `common/` dir, which covers WiFi config for all examples.  No need to add `secrets.py` into the mix.

`cheer-lights.py` should be named `cheerlights.py`. Dashes in filenames break `import` and some people like to `import example_name` from the REPL to run files.

Corteil:
I have made the suggested changes to my code and have renamed the to cheerlights.py as requested.

helgibbons:
Thanks Brian, testing it now :)

Gadgetoid:
Thanks Brian!

Note that:

```python
graphics.set_pen(color)
graphics.clear()
```

Has the same effect as:

```python
    for x in range(width):
        for y in range(height):
            graphics.set_pen(colour)
            graphics.pixel(x, y)
```

And would be much, much faster.

I do wonder if we could also do a `numpy` magic spin on this, to set the colours from the right-hand edge and propagate them across.

helgibbons:
Been running for a few hours and seems to be working good :)

helgibbons:
Looks good, many thanks!

PR #549, opened 2022-10-27
Galactic Unicorn: Add wakeup module for early button press detection.

Gadgetoid:
This probably needs to set some pull-ups at startup before reading the GPIO state?


Gadgetoid:
I can't even remember what I was doing here, but whatever it was it should probably be started again.

PR #553, opened 2022-10-29
Add time synchronization from NTP to GU clock example

MichaelBell:
Unless you're connected to Thonny, the clock will always start from midnight.

This change allows the Galactic Unicorn to connect to wifi and synchronize the time using NTP at boot, and whenever A is pressed to resynchronize.  Time from NTP is UTC, so volume up/down buttons allow the offset in hours from UTC to be set (this is retained over any later NTP synchronization).


ZodiusInfuser (approved):

helgibbons:
Looks great, thanks both :)

PR #555, opened 2022-10-30
Fix setting the LTP305 brightness

MichaelBell:
The birghtness setting on the LTP305 driver didn't work because the local variable hid the member variable.


ZodiusInfuser (approved):

helgibbons:
Thank you!

PR #565, opened 2022-11-09
Fix typo galactic_unicorn feature_test_with_audio.py

johnmccabe:
Fixes a minor typo, SWITCH_BRIGHTNESS_DOWN text should be Lower not Power.


helgibbons:
Thanks John! :)

PR #570, opened 2022-11-09
MicroPython: Enable ulab for Tiny2040 builds.

Gadgetoid:


Gadgetoid:
This probably explodes the 2MB Tiny2040. Should check that!

helgibbons:
I think the standard Pico build should work with a 2MB Tiny 2040?

Gadgetoid:
Rebased upon main, so we've got an up-to-date build to test.

PR #577, opened 2022-11-14
Badger2040: Make the LED example pulse instead of strobe

ahpook:
This makes a slightly nicer example as the LED goes gently from off to on and back down again instead of strobing.


helgibbons:
Lovely, thank you - and thanks for persevering with our fussy linting!

PR #593, opened 2022-12-03
St7789 async dma and partial update

AndrewCapon:
I put in this PR to ask for some help testing this, basically it adds async DMA transfers for the st7789 and also partial updates.

There is a test example for the pico display for testing the new code, I am also waiting for a pico display 2. When that arrives I will port over the example for that.

I do not have the other st7789 Pimoroni displays though so these will not be tested, the PIO route is also not tested.

So is this PR something you guys would like to add in, you can get a nice speedup with these changes? 

Can someone help with testing these other displays?

Cheers

Andy

Changes
=======

Added async DMA for st7789. For 565  we don’t have to wait for the SPI transfer now.

Added partial update of display so we can just update a region via SPI.

Async DMA is also implemented for partial update for 565. As the DMA on the pico doesn’t support stride this is done using two DMA channels with chained transfers.

Added ability to destruct graphics and display.

For an overview look at pico_display_async_region.cpp, this is a test example, comments at top showing what is going on.

st7789 changes
============

Constructors changed to take use_async_dma parameter, defaults to false.

command() changed to take use_async_dma flag.

is_busy() implemented, returns true if waiting for async DMA.

wait_for_update_to_finish() added, this needs to be called if using async dma. Waits for dma to finish and disables the DMA chain.

Fix for display rotating for rotate180.

Added partial_update() implementation.

setup_dma_control_if_needed() added, sets up a second DMA channel for chained transfers.

Examples changes
===============

Added example test code for the pico display: pico_display_async_region.cpp

pico_graphics changes
=====================

rect_convert_rgb565() added, like frame+convert_rgb565() but for Rects/Regions.

rect_convert() added for each pen.

create_owned_frame_buffer() added to create frame buffer owned by us

destructor added to free up frame buffer if owned.

types 
====

Rect.equals() added to test equality of Rects.

pico_graphics_pen_1bit changes
=========================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.

pico_graphics_pen_1bitY changes
==========================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.

pico_graphics_pen_3bit changes
=========================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.
Add empty rect_convert()

pico_graphics_pen_P4 changes
=========================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.
Add rect_convert()

pico_graphics_pen_P8 changes
=========================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.
Add rect_convert()

pico_graphics_pen_RGB332 changes
=============================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.
Add rect_convert()

pico_graphics_pen_RGB565 changes
=============================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.

pico_graphics_pen_RGB888 changes
=============================

Use create_owned_frame_buffer(0 to create frame buffer if none provided.

### DIFF
--- a/drivers/st7789/st7789.cpp
+++ b/drivers/st7789/st7789.cpp
@@ -110,10 +110,16 @@ namespace pimoroni {
   }
 
   void ST7789::cleanup() {
-    if(dma_channel_is_claimed(st_dma)) {
-      dma_channel_abort(st_dma);
-      dma_channel_unclaim(st_dma);
+    if(dma_channel_is_claimed(st_dma_data)) {
+      dma_channel_abort(st_dma_data);
+      dma_channel_unclaim(st_dma_data);
     }
+
+    if(use_async_dma && dma_channel_is_claimed(st_dma_control_chain)) {
+      dma_channel_abort(st_dma_control_chain);
+      dma_channel_unclaim(st_dma_control_chain);
+    }
+		
     if(spi) return; // SPI mode needs no further tear down
 
     if(pio_sm_is_claimed(parallel_pio, parallel_sm)) {
@@ -130,6 +136,10 @@ namespace pimoroni {
     if(rotate == ROTATE_90 || rotate == ROTATE_270) {
       std::swap(width, height);
     }
+		
+		// setup full and current regions
+		full_screen_region = {0, 0, width, height};
+		current_update_region = {0, 0, width, height};
 
     // 240x240 Square and Round LCD Breakouts
     if(width == 240 && height == 240) {
@@ -176,20 +186,28 @@ namespace pimoroni {
 
     // Pico Display
     if(width == 240 && height == 135) {
-      caset[0] = 40;   // 240 cols
-      caset[1] = 279;
-      raset[0] = 53;   // 135 rows
-      raset[1] = 187;
+      int col_offset = 40;
+      int row_offset = 53-rotate180; // something a little weird here, needs offsetting by -1 if rotateing 180!
+
+      caset[0] = col_offset;   
+      caset[1] = width + col_offset - 1;
+      raset[0] = row_offset;   
+      raset[1] = height + row_offset - 1;
+
       madctl = rotate180 ? MADCTL::ROW_ORDER : MADCTL::COL_ORDER;
       madctl |= MADCTL::SWAP_XY | MADCTL::SCAN_ORDER;
     }
 
     // Pico Display at 90 degree rotation
     if(width == 135 && height == 240) {
-      caset[0] = 52;   // 135 cols
-      caset[1] = 186;
-      raset[0] = 40;   // 240 rows
-      raset[1] = 279;
+      int col_offset = 52+rotate180; // something a little weird here, needs offsetting by +1 if rotateing 180!
+      int row_offset = 40; 
+
+      caset[0] = col_offset;   
+      caset[1] = width + col_offset - 1;
+      raset[0] = row_offset;   
+      raset[1] = height + row_offset - 1;
+
       madctl = rotate180 ? (MADCTL::COL_ORDER | MADCTL::ROW_ORDER) : 0;
     }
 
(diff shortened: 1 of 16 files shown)


Gadgetoid:
This is extremely interesting, thank you! Always eager to push the state of the art with our display updates and see what the Pico is capable of.

I wont get any time to feed back on this until the new year (new baby plus seemingly constant waves of sickness have demolished my productivity and I have a lot of catching up to do) but it might pique the interest of @MichaelBell who's also had some input on our display driving code.

If you want/need other screens to tinker with- not necessarily even to test or with any strings attached- let me know.

AndrewCapon:
You are always welcome to send me screens :)

The PicoDisplay2 has arrived, haven't unpacked it yet.

I am doing a ILI9341 "Pimoroni" driver at the moment, along with a XMP2046 touch screen driver. The ILI9341 has the same region and async code in, I'm going to also add new additional "pens" for direct updating via display drivers rather than via a frame buffer, this will open up larger displays to be used with the limited memory the pico has.

I don't think you sell any displays based on these but I will put in another PR for this code when I have finished the example test code. Might be useful for some of your other customers.

When I finish this I will add an example for the PicoDisplay2...

AndrewCapon:
I have added an example for the Pico Display 2 as well.

AndrewCapon:
I have added the ili9341 and xpt2046 drivers and examples : https://github.com/pimoroni/pimoroni-pico/pull/596

MichaelBell:
This looks awesome, I have also wanted to do partial updates and have some hacky patch to do that, but great to see a suggestion that could be merged.  It would be nice to expose the partial update to MicroPython.
Similarly the async DMA is a great option to have - this would give me one less reason to use my own st7789 driver for C++ projects :)

I think the extra up to ~2KB memory allocation for the chaining DMA commands could be an issue - heap RAM is at a premium especially on PicoW builds.  Maybe you could just have a single chain CB, and use the DMA complete interrupt to set it back up for each line?  Though that might occasionally stall the SPI transfer and I don't know whether the st7789 can cope with that.  That said, you haven't exposed this to MicroPython, so maybe it's not such a big deal unless that is done.

Final minor comment: the indenting is inconsistent - I think you need to convert tabs into spaces.

AndrewCapon:
Your idea of using a single chain is good if the memory is an issue.

What I would like to do is to implement a "minimal memory" display list type rendering, so you push updates to the renderer (rects, points, etc), the renderer just uses a ping pong buffer with 2 scanlines, sorts by updates by Y and intersects the objects with a single scanline which is rendered while the dma is transferring the other scanline until all updates are rendered.

So higher cpu usage but less memory, not so great for whizing things round the screen but ok for UIs. The renderer can then run on the other core. Maybe the first step is just direct to spi pens with no frame buffer at all...

What would be good is a pico with more memory!

Concerning the tabs I used the Pimoroni VSCode workspace and hoped that the tabs would be ok, I will look for some hidden setting somewhere and see what is wrong...

AndrewCapon:
I have added a pr: https://github.com/pimoroni/pimoroni-pico/pull/613 with an idea of how to use direct pens without frame buffers.

Gadgetoid:
> Your idea of using a single chain is good if the memory is an issue.

For MicroPython, memory is a huge constraint. Pretty much *any* heap allocation outside of MicroPython's own managed "heap" will either immediately, or eventually, cause an OOM lockup. This is particularly severe on Pico W where the RAM constraints are even tighter and the slightest misstep will cause the MicroPython build to not even finish booting :skull: 

Any buffers we use, must be possible to allocate using MicroPython's `m_new` or `m_malloc` which just allocate a block of its heap and gives you back a pointer. These w ill, IIRC, always be 32-bit aligned but no other alignment is guaranteed. This is why PicoGraphics is absolutely rife with buffers allocated in the MicroPython bindings and passed down the chain-

https://github.com/pimoroni/pimoroni-pico/blob/206d4089d718416dac6889f0a187bab39e094e7c/micropython/modules/picographics/picographics.cpp#L355-L368

Even our C++ classes are allocated in MicroPython's heap using `m_new_class` which is our own convenience wrapper around C++'s "Placement New"-

https://github.com/pimoroni/pimoroni-pico/blob/206d4089d718416dac6889f0a187bab39e094e7c/micropython/modules/picographics/picographics.cpp#L372-L403

This is a continuous pain point :sob:

AndrewCapon:
That is all greek to me :)

So are you saying the c/c++ code here has to be limited in order to work with the python stuff, so the python stuff is a wrapper around the c/c++ stuff and the python "runtime" uses up enough ram to limit the c/c++ stuff?

Gadgetoid:
Yep, all of the C++ stuff here (or most of it, anyway) is wrapped directly via MicroPython bindings which are just a thin veneer over the underlying C++ class to handle type conversions, argument checking, throwing errors and all that usual user facing garbage you have to deal with in a non-compiled language.

MicroPython eats RAM for breakfast, so the safest assumption is to not use any RAM the user hasn't supplied, or at least give the user the option and default to allocating if they don't supply a buffer.

We didn't know this when we started out, quite a trial by fire :grimacing:

AndrewCapon:
Ok, Got you.

So are we saying that dma_control_chain_blocks should be able to be passed in after being allocated in python or the whole idea of using the dma chaining needs to be changed?

AndrewCapon:
Ok, I just added a define USE_ASYNC_INTERRUPTS.

If this is set to 1 then the control chain is not used but interrupts are used instead.

I guess somehow you can get this define set outside of st7789.hpp depending on if using c++ or python.

Gadgetoid:
Thank you.

IIRC from @MichaelBell's musings with the camera, chained DMA requires alignment that MicroPython can't guarantee and I think that might apply here, but I'm not sure. Either way the buffers would need to be allocated by MicroPython's functions, whether those are user-supplied or handled by the bindings is relatively inconsequential.

Setting a define in the respective CMake files for MicroPython vs C++ builds should be straightforward.

I am still in very much over my head here!

MichaelBell:
The extra alignment was for use with DMA ring buffers, I haven't reviewed in detail but I don't think Andrew is doing that here.

I think I did add some explicit alignment to 4-byte boundary on the camera thing just as belt and braces, but fairly sure that buffers from Python are 4-byte aligned unless it's a user provided buffer and they've given a non-multiple of 4 offset into a larger byte buffer.

AndrewCapon:
USE_ASYNC_INTERRUPTS=1 turns chains off entirely, it is then all handled in the interrupt handler instead.

PR #602, opened 2022-12-15
Correct badge picture dimensions

housten:
Using 108 resulted in a 'smeared' image because the image width is wrong for the result.  Check the [getting started guide](https://learn.pimoroni.com/article/getting-started-with-badger-2040) where it says 104x128 for badge picture


Gadgetoid:
Good catch, thank you!

Issue #595, opened 2022-12-08
PicoSynth volume is const (and probably shouldn't be?)


ahnlak:
The otherwise excellent PicoSynth library holds a `volume`  value which is applied globally; however, this is declared as `const` which means it cannot be adjusted.

The question is whether this is deliberate (in which case, why's it a public variable on the class), or just an oversight (in which case, I have a patch for you... :smile: )

(I came across this because I realised the volume adjustment stuff in the Galactic Unicorn library doesn't actually do anything, and the rabbit hole lead me here)

PR #603, opened 2022-12-15
Unicorn volume fix

ahnlak:
In reference to [#595](https://github.com/pimoroni/pimoroni-pico/issues/595), this makes the volume in PicoSynth changeable, and then reflects the volume setting in the GalacticUnicorn library into that.

side question: should there be a distinct `volume` within GalacticUnicorn at all, or should it just fully reflect PicoSynth? I've taken the approach of touching the bare minimum, but it could be done more thoroughly if desired...

I *think* that volume is currently only `const` because the bulk of the logic has been borrowed from the 32blit synth stuff, which has an externally set volume (via the firmware) but if I'm wrong and there's a really good reason for it, feel free to torpedo this PR :smiley_cat:

### DIFF
--- a/libraries/galactic_unicorn/galactic_unicorn.cpp
+++ b/libraries/galactic_unicorn/galactic_unicorn.cpp
@@ -517,6 +517,7 @@ namespace pimoroni {
     value = value < 0.0f ? 0.0f : value;
     value = value > 1.0f ? 1.0f : value;
     this->volume = floor(value * 255.0f);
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Thank you- I'm going to take it on faith that you're probably right here, and give you this glorious merge for your troubles :laughing:  if it transpires that we want something else, it can be changed/tweaked from here.

PR #606, opened 2022-12-19
Update actions for deprecation warnings

Daft-Freak:
Just a lot of v2 -> v3.

(Removes a lot of scrolling if you're trying to get a micropython artifact)


Gadgetoid:
Thank you! I had only got as far as looking at the error output with horror.

PR #613, opened 2022-12-29
PicoGraphics direct pens

AndrewCapon:
This also contains:

https://github.com/pimoroni/pimoroni-pico/pull/593

https://github.com/pimoroni/pimoroni-pico/pull/596

The idea here is to enable PicoGraphics Pens to work without frame buffers, so directly to the display.

Display drivers need to implement the templated pure virtual interface IDirectDisplayDriver is they want to support this.

```
template<typename T> class IDirectDisplayDriver {
    public:
      virtual void write_pixel(const Point &p, T colour) = 0;
      virtual void write_pixel_span(const Point &p, uint l, T colour) = 0;
      virtual void write_pixel_rect(const Rect &r, T colour) = 0;
  };
```

`write_pixel()` should write a single pixel to the display.
`write_pixel_span()` should write a scanline/span of pixels
`write_pixel_rect()` should write a rectangle of pixels. This is currently not used.

A single direct pen has been implemented `PicoGraphics_PenRGB565_direct`

I have only updated the ili9341 driver to implement the interface, if people think this direct pen idea is a good idea I will update the ST7789 driver as well.

An example is provided `ili9341_direct_demo` has been added to show everything working.


Gadgetoid:
The direct pen concept is getting some field-testing for Inky Frame, so we'll see how that pans out. There's some discussion about whether it should be replaced with a generic hardfault handler to map the PSRAM but I'm not sure how well that would work in MicroPython.

Gadgetoid:
With PicoVision we now have a project that uses out-of-tree, non-direct display drivers - https://github.com/pimoroni/picovision

I think this is how ili9341 and xpt2046 should be handled. Our MicroPython builds are bloating out of control, and this repository is getting a little *too* monolithic to be navigable, so I'm slowly trying to break it apart into logical units and only put stuff here that's used by and shared across our products.

It looks like `IDirectDisplayDriver` and similar concepts are here to stay, though, thank you!

PR #615, opened 2023-01-02
MLX90640: Added C++ proof-of-concept ST7789/Square-LCD implementation

thinkier:
I have been eyeing this relatively inexpensive Forward-looking IR sensor for a while. Due to the Pi shortage I had been holding back. That was until I took a deep dive into the Pico ecosystem and noticed this somewhat stale branch.

# Notes
- The sensor is pretty noisy (hopefully fine for my use case of pointing at the 3D printer)
- Resolution could be an issue - FLIR Lepton would be an upgrade in this scenario. Don't think tinyUSB implements UVC though, so that will be another headache...
- The artifacting on checkerboard read pattern is pretty bad, still not as bad as interlaced mode

# Demo


Gadgetoid:
I've rebased mlx90640 and updated it to support the latest MicroPython changes.

Unfortunately the module is too big to include in our builds generally- it will overflow the RAM on a regular build and isn't super useful many in cut-down, board-specific builds. It feels like we might need a pure MicroPython approach here for it to have much mileage.

PR #628, opened 2023-01-10
Use the LUX sensor to automatically set brightness

LionsPhil:
Combine the board features in the suggested fashion in the most featureful badge example. Read the photosensor and set the display brightness accordingly.

### DIFF
--- a/micropython/examples/tufty2040/retro_badge.py
+++ b/micropython/examples/tufty2040/retro_badge.py
@@ -1,6 +1,7 @@
 # A retro badge with photo and QR code.
 # Copy your image to your Tufty alongside this example - it should be a 120 x 120 jpg.
 
+from machine import ADC, Pin
 from picographics import PicoGraphics, DISPLAY_TUFTY_2040
 from pimoroni import Button
 import time
@@ -9,6 +10,9 @@ import qrcode
 
 display = PicoGraphics(display=DISPLAY_TUFTY_2040)
 button_c = Button(9, invert=False)
+lux_pwr = Pin(27, Pin.OUT)
+lux_pwr.value(1)
+lux = ADC(26)
 
 WIDTH, HEIGHT = display.get_bounds()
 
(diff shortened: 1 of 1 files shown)


LionsPhil:
(This is tested on my Tufty2040, since I don't see an automatic build test for that.)

ahnlak:
Looks like it doesn't like your line break in 
```
   display.set_backlight(BACKLIGHT_LOW +
        (luminance_frac * (BACKLIGHT_HIGH - BACKLIGHT_LOW)))
```

LionsPhil:
Weird, I *don't* get that finding locally. Looks like this check is also failing on other contempoary PRs in the same way, so I assume it's an infra problem and leave the commits as-is for now.

Gadgetoid:
We're linting with `flake8` which may have a difference of opinion with pylint.

Run with:

```
flake8 --show-source --ignore E501
```

LionsPhil:
There we go:
```
micropython/examples/tufty2040/retro_badge.py:155:41: W504 line break after binary operator
    display.set_backlight(BACKLIGHT_LOW +
                                        ^
micropython/examples/tufty2040/retro_badge.py:156:9: E128 continuation line under-indented for visual indent
        (luminance_frac * (BACKLIGHT_HIGH - BACKLIGHT_LOW)))
        ^
```
Line rewrapped in a way it's happy with.
(Can you bug the premerge check not giving this output?)

Gadgetoid:
@helgibbons dare I ask if you managed to test this :eyes:

helgibbons:
This (and #629 ) works - thanks @LionsPhil , and super-thanks for fixing the Tufty launcher!  However, I'm finding it quite flickery at certain brightness levels. I think the LED lights in this office may be the culprit, as I've just wandered round the building and under different kinds of lights it's absolutely fine. Could be possible to add more averaging/sampling to improve this behaviour, perhaps?

It might be a good idea to have the brightness/battery adjusting version as a separate example due to the added complexity, but if we are adding it to the default example it might be best to have this functionality turned off by default. I really like the dimming being connected to both light level and battery level though :)

helgibbons:
Here's a video which hopefully shows the flickering a little:
https://user-images.githubusercontent.com/50950368/221248391-9ad2dafe-6268-4812-9207-812bfbeb7c71.mp4

LionsPhil:
Ah yeah, that's annoying. Alright, my actual personal badge example has a combination of these which iterates on them a bunch, including some smoothing; I'll see if I can pull that out as a standalone example. I assume it's cleanest to make that a separate PR and retract these two?

Gadgetoid:
Replaced with #701

PR #629, opened 2023-01-10
Limit display brightness on low battery.

LionsPhil:
Use another feature of the board in the featureful demo: measure battery voltage while on battery, and set minimum brightness if it is low.

This shares a commit with #628, but adds a more potentially-contested (and unfortunately less tested) feature, so is a separate PR. I don't know if there is any significant battery lifespan hit from constantly reading the voltage in a tight loop. (I *have* tested with an artificially high voltage threshold that it works on my Tufty2040, but don't have a bench supply or anything to really do that thoroughly, transitioning from high to low.)


LionsPhil:
(Will wait for #628 to close before fixing/redoing/aborting this one.)

Gadgetoid:
Replaced with #701

PR #632, opened 2023-01-11
Update network_manager.py

ben4github:
In order to successfully connect to more complicated AP setups (ie WPA-WPA2-WPA3) mesh networks, the power save mode needs to be in place. It is possible that a different mode can be set once a socket is opened. For now this is necessary for a range of ASUS routers/APs to be able to negotiate a connection.

### DIFF
--- a/micropython/examples/common/network_manager.py
+++ b/micropython/examples/common/network_manager.py
@@ -23,6 +23,9 @@ class NetworkManager:
         return self._sta_if.isconnected() or self._ap_if.isconnected()
 
     def config(self, var):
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you.

I don't think this is the right place to put it in `network_manager`, since getting config vars should not have a side effect.

It should probably go just above the connect line, here - https://github.com/pimoroni/pimoroni-pico/blob/f17c04b7ffed670002df9456926cedaef2bb8149/micropython/examples/common/network_manager.py#L77

Gadgetoid:
Thank you, I've rolled this into https://github.com/pimoroni/pimoroni-pico/pull/731

PR #650, opened 2023-01-25
EXPERIMENT: Grab custom GCC toolchain from GitHub actions.

Gadgetoid:
This is simple, really, another repository - https://github.com/Gadgetoid/arm-gnu-toolchain-build - builds us a nice tweaked version of the ARM GNU toolchain with the `--disable-libstdcxx-verbose` that's used in Ubuntu packaged versions of the toolchain but not those available from ARM's website.

We grab that tweaked GCC, unpack it (twice :roll_eyes:) and bung the bin dir into the PATH.

Aaand hey presto we're building against a - hopefully - working toolchain that we can also run on our local machines without requiring Ubuntu 20.04 or exploding our systems with crusty deb packages.

Something something docker something something.  :shushing_face:


Gadgetoid:
Okay, first attempt to actually *run* a .uf2 from this build leads to "oh no, it does not work :sob:"

Something... has gone wrong.

Gadgetoid:
After the MicroPython greedy heap change, I seem to be producing builds that run locally. If I revert that change, they are bloated by the size of the "gc_heap" variable and inevitably this pushes them over the limit. This is... odd... since I explicitly marked that variable as uninitialized here: https://github.com/micropython/micropython/commit/71f6eb5ac953637e815ec9c5cbc785b4027a8759

Possible that for some reason this compiler version is not respecting the `NOLOAD` in the linker script?

Either way, I need to rebase this and see what happens because it might produce at least running builds now.

PR #653, opened 2023-01-27
Badger W: partial updates for clock

MichaelBell:
I was just looking through the Badger examples and thought the clock would look quite a lot nicer with some partial updates.

I've also slowed down the per-minute update to make it stick a bit better.


Gadgetoid:
You know, I had totally forgotten that partial updates existed. It works really nicely, although the rest of the time does fade pretty quickly. It's almost a feature, though.

Thank you!

PR #655, opened 2023-01-30
i75/i75w: Auto-detect board and set up user button.

ageeandakeyboard:
Allows for the Interstate75 (non W) user button to be used with our new python module

### DIFF
--- a/libraries/interstate75/README.md
+++ b/libraries/interstate75/README.md
@@ -27,6 +27,9 @@ PicoGraphics_PenRGB888 graphics(hub75.width, hub75.height, nullptr);
 
 // And each button
 Button button_a(Interstate75::A);
+// For the Interstate75
+Button button_b(Interstate75::BOOT);
+// Or for the Interstate75W
 Button button_b(Interstate75::B);
 
 // RGB LED
--- a/micropython/modules_py/interstate75.md
+++ b/micropython/modules_py/interstate75.md
@@ -27,7 +27,7 @@ DISPLAY_INTERSTATE75_192X64
 DISPLAY_INTERSTATE75_256X64
 ```
 
-You are able to set which version of the Interstate75 that you are using with the `interstate75w=` as default this value is set to `False` to use the Interstate75 button layout by setting it to `True` the module will use the Intersate75W button layout. If you are using an Interstate75W it is important to set this value to `interstate75w=True`, otherwise this could disable the wifi in your project.
+You are able to set which version of the Interstate75 that you are using with the `interstate75w=`. By default this value is set to `True` to use the Interstate75W button layout. By setting it to `False` the module will use the Interstate75 button layout. If you are using an Interstate75W it is important to set this value to interstate75w=True, otherwise this could disable the WiFi in your project
 
 ```python
 import interstate75
@@ -44,7 +44,7 @@ import interstate75
 
 display = interstate75.DISPLAY_INTERSTATE75_32X32
 
-board = interstate75.Interstate75(display=display, interstate75w=True)
+board = interstate75.Interstate75(display=display, interstate75w=False)  #This is th e setup to be able to access the both A and BOOT on the Interstate75 (non W)
 ```
 
 From here, all features can be accessed by calling functions on `board`. In addition, when using Qwiic / Stemma QT devices, the I2C channel to use can be accessed with `board.i2c`.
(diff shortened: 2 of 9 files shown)


ZodiusInfuser:
Thanks @ageeandakeyboard. This should really do some handling of Interstate75 vs Interstate75W. GPIO23 which is used for boot button in Interstate, is the WIFI_ON pin on Interstate W, so we should not be touching that.

ZodiusInfuser:
https://github.com/pimoroni/pimoroni-pico/tree/main/libraries/interstate75

micropython/modules_py/interstate75.py
```diff
@@ -55,7 +61,7 @@ def update(self, buffer=None):
 
     def switch_pressed(self, switch):
         if switch < 0 or switch >= self.NUM_SWITCHES:
-            raise ValueError("switch out of range. Expected SWITCH_A (0), SWITCH_B (1)")
+            raise ValueError("switch out of range. Expected SWITCH_A (0), SWITCH_B (1) SWITCH_BOOT (2)")
```
ZodiusInfuser, line 64: This should go back to how it was before, though with SWITCH_B/BOOT

micropython/modules_py/interstate75.md
```diff
@@ -27,6 +27,7 @@ DISPLAY_INTERSTATE75_192X64
 DISPLAY_INTERSTATE75_256X64
 ```
 
+You are able to set which version of the Interstate75 that you are using with the `interstate75w=` as default this value is set to `False` to use the Interstate75 button layout by setting it to `True` the module will use the Intersate75W button layout. If you are using an Interstate75W it is important to set this value to `interstate75w=True`, otherwise this could disable the wifi in your project.
```
ZodiusInfuser, line 30: Suggested edit:
"You are able to set which version of the Interstate75 that you are using with the `interstate75w=`. By default this value is set to `False` to use the Interstate75 button layout. By setting it to `True` the module will use the Interstate75W button layout. If you are using an Interstate75W it is important to set this value to `interstate75w=True`, otherwise this could disable the WiFi in your project."

ZodiusInfuser:
Thanks. I suggested some edits. It could also do with the updates to the C++ definitions and readme too: https://github.com/pimoroni/pimoroni-pico/tree/main/libraries/interstate75 i.e. the addition of `static const uint8_t BOOT = 23;` to the header

libraries/interstate75/README.md
```diff
@@ -27,6 +27,9 @@ PicoGraphics_PenRGB888 graphics(hub75.width, hub75.height, nullptr);
 
 // And each button
 Button button_a(Interstate75::A);
+// For the Interstate75
+Button button_b(Interstate75::BOOT);
```
ZodiusInfuser, line 31: Should this be commented out?

micropython/modules_py/interstate75.md
```diff
@@ -44,7 +44,7 @@ import interstate75
 
 display = interstate75.DISPLAY_INTERSTATE75_32X32
 
-board = interstate75.Interstate75(display=display, interstate75w=True)
+board = interstate75.Interstate75(display=display, interstate75w=False)  #This is th e setup to be able to access the both A and BOOT on the Interstate75 (non W)
```
ZodiusInfuser, line 47: "th e" should be "the"

ZodiusInfuser: Oh and a space is needed after the #, otherwise linting will shout

libraries/interstate75/README.md
ZodiusInfuser: Oh, it's the readme

micropython/modules_py/interstate75.md
ageeandakeyboard: Opps slip of the keyboard :)

Gadgetoid:
There must be a way to auto-detect that the code is running on a Pico W, rather than relying upon user testimony?

ZodiusInfuser:
Since we have Pico W and Non-W builds, we could be evil and swap out which version of interstate75 gets baked in.

ageeandakeyboard:
It makes sense but then we i'll need to have two differently named modules.

Gadgetoid:
I think the user-facing logic here can be replaced with `"Pico W" in sys.implementation._machine`, I might grab/rebase this with that change to keep it fresh.

Gadgetoid:
@ageeandakeyboard if you still have i75 and i75w boards to hand, could you see if this works please?

PR #616, opened 2023-01-02
Unicorn rgb332 fix

ahnlak:
Enables support for PEN_RGB332 on the Galactic Unicorn; this is necessary to support the sprite functions of PicoGraphics, which only work with that pen.

PR #659, opened 2023-01-31
Added PEN_RGB332 support

ahnlak:
Enables support for PEN_RGB332 on the Galactic Unicorn; this is necessary to support the sprite functions of PicoGraphics, which only work with that pen.

(This replaces https://github.com/pimoroni/pimoroni-pico/pull/616 with a single commit PR that doesn't mess up random other line endings too, after I made a *real* mess of that original branch trying to make it better - sorry!)


Gadgetoid:
Thank you for jumping through the hoops to clean it up! (and for the fix!)

PR #666, opened 2023-02-10
Fix ghosting on the Galactic Unicorn

MichaelBell:
Select the row before streaming out the pixel data, as the row change takes a few cycles to propagate.


Gadgetoid:
Awesome stuff, thank you!

I guess I'd better get a release out :grimacing:

PR #673, opened 2023-02-20
Update to Badger 2040 W getting started

Zoobdude:
When installing the alternative firmware I saw the notice at the top of the README that the "badger2040w.py" and "network_manager.py" files need to be downloaded separately. After adding these two scrips to the Badgerw it gave `ImportError: no module named 'WIFI_CONFIG'`. This was not something that the README said was required. Therefore my PR adds the "WIFI_CONFIG.py" to the top README file.


helgibbons:
Thanks for this - think we'd all got used to grabbing the -with-examples version :)

PR #675, opened 2023-02-21
Refresh rate and raw

iMartyn:
Does this update the python side? who knows, let's see.

### DIFF
--- a/drivers/hub75/hub75.cpp
+++ b/drivers/hub75/hub75.cpp
@@ -6,7 +6,7 @@
 
 namespace pimoroni {
 
-Hub75::Hub75(uint width, uint height, Pixel *buffer, PanelType panel_type, bool inverted_stb, COLOR_ORDER color_order, REFRESH_RATE refresh_rate)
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I'm not sure what you're trying to accomplish here, do you have an FM6126A based panel that has a different way to set up the outputs on the shift registers?

It might be easier to hack on either RPi's original HUB75 demo code, or my hello world C++ code which is much more down to earth and thoroughly documented, see here: https://github.com/pimoroni/pimoroni-pico/blob/main/examples/interstate75_legacy/interstate75_hello_world.cpp

Once you've figured out what the sekret sauce is for your particular panel, we can see if it's possible to reconcile that with this driver or if it's too weird to be worth the compromise.

Either way, this doesn't need to be a PR since CI should run on your local copy and produce builds - https://github.com/iMartyn/pimoroni-pico/actions

iMartyn:
Yeah, sorry, I was attempting to create a PR on my fork, didn't spot it'd crossed forks.

PR #679, opened 2023-02-21
Update function names in readme

Zoobdude:
Looks like someone forgot to update the function names in the Badgerw's readme (they were still the ones from the old library).


helgibbons:
Good spot, thank you!

PR #695, opened 2023-02-27
Launcher for Inky Frame 7.3"

thirdr:
- Adds Inky Frame 7.3" support to the Launcher


Gadgetoid:
(https://soundcloud.com/xbox/achievement-mp3-sound)

:trophy: Achievement Unlocked: Use the --force.

PR #701, opened 2023-03-02
Contribute an automatic display brightness example

LionsPhil:
Standalone smoothed display auto-brightness, reactive to the on-board LUX and battery ADC sensors, with debugging keys so it's easier to test (and play with).

Closes pull requests #628 and #629, which did this to the retro_badge example, but made it too complicated.


helgibbons:
This is great, many thanks for your efforts!

PR #708, opened 2023-03-06
Fix small typo in README

antonmosich:


helgibbons:
Good spot, thanks very much!

PR #711, opened 2023-03-10
MicroPython: Avoid heap allocations in all C++ modules. 

Gadgetoid:
As of https://github.com/micropython/micropython/commit/c80e7c14e6305e50e3b39f97172d4d8fe1214d3b MicroPython's linker script includes a "greedy" heap, which attempts to use *all* free system RAM for MicroPython's heap.

Whereas previously we carefully avoided large heap allocations due to the very limited amount of heap available to C++ extensions, we must now avoid *all* allocations.

This changeset cleans up some uses of `std::string` which were implicitly allocating strings on heap. It does this by switching to `string_view`, wrapping the existing byte array and length returned by `GET_STR_DATA_LEN` instead of *copying* them into a `std::string`.

Additionally `set_font()` performs Hershey font lookups via a new `if/else` table, since the `std::map<std::string, const *font_t>` lookup table could not be used without constructing a string and risking a heap allocation.

Finally I have removed the completely ridiculous and unnecessary conversion to/from a `std::string` in our jpegdec binding's `open_file` functionality.

To make `string_view` available, all MicroPython builds needed to be forced to use C++17, this is achieved in their respective CMake files.

In all cases I believe `std::string` will implicitly cast to `std::string_view` so our examples will not break, and C++ code has always been set to C++17.


Gadgetoid:
This is a pre-requisite to bumping all of our MicroPython builds to the latest edge commit and - indeed - keeping ahead of MicroPython changes so we can incorporate Bluetooth when it arrives.

Once merged, relevant commits from https://github.com/pimoroni/badger2040/tree/test/micropython-edge can be moved to https://github.com/pimoroni/badger2040/pull/2 (and the GitHub actions .yml targeted back at the main branch here)

Gadgetoid:
Pico RGB Keypad, Pico Scroll and Pico Unicorn have all been converted to MicroPython class-style modules. `init()` changes to a new class instance. For example this:

```python
import picounicorn

picounicorn.init()
```

Would become:

```python
from picounicorn import PicoUnicorn

picounicorn = PicoUnicorn()
```

Gadgetoid:
Looks like `interstate75.py` is not included in Pico W builds and should be. The MicroPython module .cmake files need tidying *again* since they've got quite out of hand.

Issue #578, opened 2022-11-15
clock.py - utc_offset errors


n21brown:
pressing Vol up/down increments the utc_offset by more than one each press.  As a result this allows utc_offset values to greater than 24 which then shows times greater than 23:59:59

helgibbons:
Should be improved with #715 :)

PR #715, opened 2023-03-15
clock.py: constrain hour to 0 to 24

mogenson:
Use a modulo to constrain the displayed hour to 0 - 24 after applying the UTC offset.

Very small change to the Galactic Unicorn, Cosmic Unicorn, and Interstate75 `clock.py` examples.


helgibbons:
Looks good, thank you!

I think this may help fix #578 ?

PR #722, opened 2023-03-21
Badger 2040: Add deprecation warnings and firmware links.

Gadgetoid:


Gadgetoid:
Phase 2 will be to delete all the deprecated content from this repository, but currently the Badger 2040 repository isn't quite ready.

Issue #720, opened 2023-03-19
Pico Display 2.0 Button Test


rsrobbins:
The button test example for the Pico Display does not appear to work for the Pico Display 2.0. I changed DISPLAY_PICO_DISPLAY to DISPLAY_PICO_DISPLAY_2. None of the buttons work and nothing else happens.

helgibbons:
The code below seems to be working as expected for me using the latest release - which UF2 are you using (and what board are you using it on?)

```
# This example shows you a simple, non-interrupt way of reading Pico Display's buttons with a loop that checks to see if buttons are pressed.

import time
from pimoroni import Button
from picographics import PicoGraphics, DISPLAY_PICO_DISPLAY_2, PEN_P4

# We're only using a few colours so we can use a 4 bit/16 colour palette and save RAM!
display = PicoGraphics(display=DISPLAY_PICO_DISPLAY_2, pen_type=PEN_P4, rotate=0)

display.set_backlight(0.5)
display.set_font("bitmap8")

button_a = Button(12)
button_b = Button(13)
button_x = Button(14)
button_y = Button(15)

WHITE = display.create_pen(255, 255, 255)
BLACK = display.create_pen(0, 0, 0)
CYAN = display.create_pen(0, 255, 255)
MAGENTA = display.create_pen(255, 0, 255)
YELLOW = display.create_pen(255, 255, 0)
GREEN = display.create_pen(0, 255, 0)


# sets up a handy function we can call to clear the screen
def clear():
    display.set_pen(BLACK)
    display.clear()
    display.update()


# set up
clear()

while True:
    if button_a.read():                                   # if a button press is detected then...
        clear()                                           # clear to black
        display.set_pen(WHITE)                            # change the pen colour
        display.text("Button A pressed", 10, 10, 240, 4)  # display some text on the screen
        display.update()                                  # update the display
        time.sleep(1)                                     # pause for a sec
        clear()                                           # clear to black again
    elif button_b.read():
        clear()
        display.set_pen(CYAN)
        display.text("Button B pressed", 10, 10, 240, 4)
        display.update()
        time.sleep(1)
        clear()
    elif button_x.read():
        clear()
        display.set_pen(MAGENTA)
        display.text("Button X pressed", 10, 10, 240, 4)
        display.update()
        time.sleep(1)
        clear()
    elif button_y.read():
        clear()
        display.set_pen(YELLOW)
        display.text("Button Y pressed", 10, 10, 240, 4)
[...]

PR #725, opened 2023-03-23
MicroPython: Magic C++ Memory Management.

Gadgetoid:
This is more a test to see if I'm barking up the right tree than a fix... yet.

### DIFF
--- a/micropython/modules/badger2040-micropython.cmake
+++ b/micropython/modules/badger2040-micropython.cmake
@@ -21,3 +21,6 @@ endif()
 include(plasma/micropython)
 include(ulab/code/micropython)
 include(qrcode/micropython/micropython)
+
+# C++ Magic Memory
+include(cppmem/micropython)
--- a/micropython/modules/cppmem/cppmem.c
+++ b/micropython/modules/cppmem/cppmem.c
@@ -1,0 +1,30 @@
+#include "cppmem.h"
+
+STATIC MP_DEFINE_CONST_FUN_OBJ_0(cpp_mem_get_alloc_bytes_obj, cpp_mem_get_alloc_bytes);
+STATIC MP_DEFINE_CONST_FUN_OBJ_0(cpp_mem_get_alloc_count_obj, cpp_mem_get_alloc_count);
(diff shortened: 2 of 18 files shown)


Gadgetoid:
Okay as a result of sleuthing for #720 I ended up raising https://github.com/micropython/micropython/issues/11116

After raising my concerns in https://github.com/micropython/micropython/commit/c80e7c14e6305e50e3b39f97172d4d8fe1214d3b

And spending a large amount of time trying to *avoid* `new` and `delete`.

Unfortunately, C++ does a lot of heap allocation that's difficult to control. In fact our project does ~10k and it's a job for Future Phil to figure out why on earth we're using *so much* heap because that seems extreme.

As a result of hours upon hours of tinkering I finally came up with https://github.com/Gadgetoid/micropython-cppmem which is integrated directly into pimoroni-pico in this PR.

This new module just needs to be linked into a MicroPython build and it takes over all C++ memory allocations in two ways:

1. All pre-init allocations end up in a 10k, linear, non-freeing memory pool that survives until hard-reset
2. Once init is done, the module switches into "MICROPYTHON" mode, where `new` and `delete` map to `m_malloc` and `m_free` respectively.

A side-effect of this approach is that any memory resulting from an implicit or explicit use of `new` in C++ modules *cannot be relied upon to not be garbage collected.*

This is to say that all our best practices still apply- if you're allocating something that you want to last for more than one C++ function call then it needs to be allocated with `m_new` and a pointer to the resulting memory region must be stored on an active MicroPython object. Like how our classes work with pointers to buffers and i2c objects.

Gadgetoid:
The builds breaking and failing to start up when the `cppmem/micropython` module is included *before other modules* suggest that `__attribute__ ((init_priority (65535)))` is *not* global. If this is the case, then moving the wakeup module further up the list of C++ modules could make Badger 2040 / Badger 2040 W wakeup button detection a little more responsive.

Gadgetoid:
Should probably move the switch to MicroPython heap mode into `boot.py` since we can be sure it will *always* run after the MicroPython heap, and C++ objects, have initialised.

PR #735, opened 2023-03-28
CI: Experimental: Set nano.specs on MicroPython C++ root module.

Gadgetoid:
PR opened just to surface CI and this problem in general.

I was unable to account for around 8512 bytes of RAM in my local MicroPython builds. IE: We were not seeing the large jump in free RAM on MicroPython's heap as indicated by https://github.com/micropython/micropython/commit/c80e7c14e6305e50e3b39f97172d4d8fe1214d3b

Some probing revealed this usage to be related to C++ stack unwinding and exception handling which, despite trying to turn it off, wouldn't strip from the build without using `nano.specs`. Even an empty `USER_C_MODULES` .cmake file would trigger this.

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -7,7 +7,7 @@ on:
     types: [created]
 
 env:
-  MICROPYTHON_VERSION: 38e7b842c6bc8122753cbf0845eb141f28fbcb72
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Looks like this frees about 10,880 bytes of RAM that MicroPython's heap will use instead.

Have raised the required MicroPython changes in https://github.com/micropython/micropython/pull/11143

PR #737, opened 2023-03-28
CI: Tidyup and combine MicroPython builds.

Gadgetoid:


Gadgetoid:
This is getting... somewhere. I couldn't help but come back for another crack at this.

The TLDR is- our build setup wasn't exactly robust and up to date, and this should - with any luck - give us a more rational set of steps that aren't against the grain and hard to debug.

Mostly-

* Stop *copying* board dirs into the MicroPython and Pico SDK source trees :grimacing: 
* Use `-DMICROPY_BOARD_DIR` instead
* Borrow a trick to make the `PIMORONI_TUFTY2040` board let Pico SDK correctly locate `pimoroni_tufty2040.h`
* Explicitly denote which boards need hacks and local board dirs in the workflow matrix- this shows up in CI runs, so we can tell at a glance with flavour of build it is
* rename *everything* so it's not so darn wordy, no more `MicroPython / Build picow_cosmic_unicorn`, it's now `MicroPython / cosmic_unicorn`

Issue #647, opened 2023-01-22
Wanted: Galactic Unicorn Wireless C++ example+doc


flavio-fernandes:
I'm interested in using the wireless functionality with the Galactic Unicorn.
The documentation and examples for doing that seem very terse at the moment.
Ideally, I would like to write a C++ program that would leverage the library
https://github.com/pimoroni/pimoroni-pico/tree/main/libraries/galactic_unicorn 

Possible? The closest I got to that is `Galactic-Unicorn-MQTT-Scroller`:
https://github.com/ucl-casa-ce/Galactic-Unicorn-MQTT-Scroller/tree/main/micropython
But I could not get that to work, and would rather not use python if possible.

PR #738, opened 2023-03-28
Added an HTTP example for the Galactic Unicorn

ahnlak:
Inspired somewhat by https://github.com/pimoroni/pimoroni-pico/issues/647, I've put together a basic "grab a quote from the Internet and display it on the Unicorn" type example.

I've endeavoured  to keep it close in format to the other examples, and commented the code to hopefully make it intelligible to others (or at least, as intelligible as it's possible to be with lwIP code!)

You need to include `-DPICO_BOARD=pico_w` to `cmake` in order to be able to pick up the WiFI-related libraries.

You also need to include `-DWIFI_SSID="ssid"` and `-DWIFI_PASSWORD="password"` to provide suitable WiFi credentials (although I do have a functioning example of providing this through a mounted USB drive, it's ... a much bigger bit of work!)

Hopefully of some interest?!

### DIFF
--- a/examples/galactic_unicorn/CMakeLists.txt
+++ b/examples/galactic_unicorn/CMakeLists.txt
@@ -123,3 +123,39 @@ pico_enable_stdio_usb(feature_test 1)
 
 # create map/bin/hex file etc.
 pico_add_extra_outputs(feature_test)
+
+# WiFi / Bluetooth examples, only built when the board type is set to PicoW
+if (${PICO_BOARD} STREQUAL "pico_w")
+
+  add_executable(
+    scrolling_quote
+    scrolling_quote.cpp
+  )
+
+  # Pull in the pico libraries that we need
+  target_link_libraries(
+    scrolling_quote
+    pico_stdlib pico_graphics galactic_unicorn
+    pico_cyw43_arch_lwip_threadsafe_background pico_lwip_http
+  )
+
+  # Need to include lwipopts.h from the local dir
+  target_include_directories(
+    scrolling_quote PRIVATE ${CMAKE_CURRENT_LIST_DIR}
+  )
+
+  # WiFi connection details; provide these to your CMake command!
+  target_compile_definitions(
+    scrolling_quote PRIVATE
+    WIFI_SSID=\"${WIFI_SSID}\"
+    WIFI_PASSWORD=\"${WIFI_PASSWORD}\"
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Of great interest, thank you!

I'm actually hoping to give both Galactic and Cosmic Unicorn the same treatment as Badger 2040 - ie: moving them and their examples to another GitHub repository where firmware, examples and documentation will be easier to find.

As such I'm not in a tremendous hurry to merge this, but it will be very valuable as a test case for the new repository.

ahnlak:
Makes a lot of sense for the Unicorns to get the Badger treatment.

Although, on reflection this stuff is PicoW-related rather than Unicorn specific, so I'm wondering if there's a more rational place for it (and also the USB Mass Storage thing if I ever lick it into a comprehensible state!)?

Gadgetoid:
Probably wouldn't hurt you to throw it in its own repo using our Pico Boilerplate, and we can link to it from there?

If you get the FAT stuff working you can ship a .uf2 and we can link it from the product page(s) too. Since I guess that means the user can config WiFi over USB?

As much as I *love* contributions, stuff tends to get absolutely lost in this havoc of a repository and I'd rather projects like yours thrive and get some attention rather than die in a cold, dark corner with a billion other examples.

ahnlak:
That's precisely what set me off down that rabbit hole, yeah - my "improved" [Unicorn clock](https://github.com/ahnlak/uniclock) does exactly that; gives you a `CONFIG.TXT` to edit to get the WiFi playing nicely.

I need to turn that into boilerplate-y thing for future projects too (hmm, can boilerplates be based on boilerplates?) and it probably makes sense to roll some HTTP stuff into that given that it's all very WiFi-centric.

Will ping you when it's fit for consumption :)

ahnlak:
Quick follow-up to this; I've rebuilt this example into it's own repo at https://github.com/ahnlak-rp2040/gu-scrolling-quote - it's based off a more comrehensive PicoW boilerplate that bakes in the filesystem and a simple http(s) client :-)

Gadgetoid:
Adding some links in the README via this PR: https://github.com/pimoroni/unicorn/pull/20

Unicorn Python examples currently live here: https://github.com/pimoroni/unicorn

No home yet for C++ stuff, which could be in this same repo or just a collection of stand-alone demo projects and a boilerplate.

PR #746, opened 2023-04-08
Fix web server only serving file once and wifi country configuration

mutatrum:
The Galactic Unicorn example code html_text.py only serves the index.html file only once.

Secondary, the WIFI country code is defined in the NetworkManager code.

### DIFF
--- a/micropython/examples/automation2040w/web_io_interface/main.py
+++ b/micropython/examples/automation2040w/web_io_interface/main.py
@@ -39,9 +39,7 @@ try:
     from tinyweb.server import webserver
 
 except ImportError:
-    # WIFI settings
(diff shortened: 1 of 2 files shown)


mutatrum:
Seems like an unrelated build failure? Can it be triggered again?

Gadgetoid:
Actually this is just a Python change so there's no reason the C++/MicroPython CI needs to run at all.

Looks good to me- and I see exactly why the old `read()` approach doesn't work :grimacing:

mutatrum:
Thanks for the merge, hope it helps future people. It definitely stumped me for a while when trying out the Galactic Unicorn.

PR #747, opened 2023-04-09
Inky Frame word clock example - swapping out NTP for Open Meteo API

martin-hamilton:
This PR swaps out the NTP call  for the word clock app and replaces it with the Badger 2040W weather app code which calls the Open Meteo API, then it tries to set the RTC based on the results of the Open Meteo API call. With a following wind we should end up with the correct local time for the lat/long supplied, rather than UTC. Just a quick hack so not expecting you to commit to the repo, but might be useful somewhere!

### DIFF
--- a/micropython/examples/inky_frame/inkylauncher/word_clock.py
+++ b/micropython/examples/inky_frame/inkylauncher/word_clock.py
@@ -1,5 +1,11 @@
 import machine
-import ntptime
+import urequests
+
+# Set your latitude/longitude here (find yours by right clicking in Google Maps!)
+LAT = 53.38609085276884
+LNG = -1.4239983439328177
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you, currently blocked by a couple of linting issues- the CI logs have gone, but:

```
flake8 micropython/examples/inky_frame/inkylauncher/word_clock.py  --ignore E501
micropython/examples/inky_frame/inkylauncher/word_clock.py:43:1: W293 blank line contains whitespace
micropython/examples/inky_frame/inkylauncher/word_clock.py:61:5: F841 local variable 'e' is assigned to but never used
```

Gadgetoid:
The MicroPython build and examples for Inky Frame are moving here: https://github.com/pimoroni/inky-frame/tree/main

PR #751, opened 2023-04-13
Update noise.py

phennessey7:
Updated the song notes to play a more accurate version of the "Sailor's Hornpipe" tune!


helgibbons:
Thanks very much for this :)

Issue #744, opened 2023-04-06
Inky Frame sleep_for >= 1 day?


ahnlak:
Looking at the code in https://github.com/pimoroni/pimoroni-pico/blob/faf4efac34f0bbc54561d68de82ad72ffde4b950/micropython/modules_py/inky_frame.py#L82
it seems that sleeps over 24 hours aren't handled (in fact, things get increasingly weird if it's more than 2 days...!) - before I go and pull a patch together, is there a good reason (does the RTC or alarms grow unreliable beyond 24 hours?) or is it just a simple oversight (or assumption nobody would be sleeping that long!)

If not, I'm happy to send a suitable PR (slightly complicated by the fact I don't actually own any Inkys so testing is ... nontrivial!)

PR #754, opened 2023-04-21
Inky Frame: longer sleep_for

ahnlak:
Expanded `sleep_for()` function to allow for sleeps longer than 24 hours - by adding the day component to the alarm, this maximum can be stretched to a month (capped at 28 days)

I've validated the code as best I can, but I don't actually possess an Inky Frame myself :-)

ref: https://github.com/pimoroni/pimoroni-pico/issues/744

### DIFF
--- a/micropython/modules_py/inky_frame.py
+++ b/micropython/modules_py/inky_frame.py
@@ -87,17 +87,21 @@ def sleep_for(minutes):
     if second >= 55:
         minute += 1
 
-    minute += minutes
+    # Can't sleep beyond a month, so clamp the sleep to a 28 day maximum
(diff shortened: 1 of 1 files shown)


ahnlak:
I saw all those errors and decided they *probably* weren't my fault :)

Gadgetoid:
Hoookay, I think I've found the cache bugs. If you can rebase this it *might* just pass CI :sob:

Gadgetoid:
Hooray! This should also work for Badger 2040 W, which has the same function here: https://github.com/pimoroni/badger2040/blob/main/firmware/PIMORONI_BADGER2040W/lib/badger2040.py#L133-L154

PR #764, opened 2023-04-28
Reduce size & workload of MicroPython build cache

Gadgetoid:
By cloning only the modules we actually intend to use, it should be slightly quicker and result in a smaller cache file for each individual board build.


Gadgetoid:
Fetching submodules using "make submodules" seems to behave differently locally vs GitHub, resulting in missing modules in the MicroPython cache.

This whole cache malarkey is almost more trouble than it's worth, but cloning MicroPython and its submodules ten times per workflow run would be quite miserable.

Have manually added the lines to clone each required submodule.

Takes the MicroPython cache from ~879 MB down to ~187 MB for a total reduction of around 6,920 MB across all builds :grimacing:

PR #771, opened 2023-05-08
Fixes for compilation under macOS / GCC 13.3.1

raybellis:
These fixes were required to make the current `main` branch compile on macOS 13.3.1 using `arm-none-eabi-gcc @13.1.0` from MacPorts.


Gadgetoid:
Thank you!

PR #774, opened 2023-05-10
Support for RGB Encoder Wheel Breakout

ZodiusInfuser:


ZodiusInfuser:
This is now ready for review

Gadgetoid:
Found *one* typo, but I think I might handle that in another PR because there are one or two more across our codebase :laughing: 

I guess it's time for a new release. Should be the last to include Cosmic/Galactic builds, too.

PR #775, opened 2023-05-10
Ghosting fix for Cosmic

MichaelBell:
This is applying the same ghosting fix as for Galactic to Cosmic.  I've just given it a brief test and it seems to be working.

The change is to apply the row selection earlier, as it appears to take a few cycles to propagate so the previous method of changing it immediately before display starts very briefly lights the previous row.


Gadgetoid:
Oooh, thanks for picking up my slack here. I could not wrap my head around the necessary changes.

I have no earthly clue what's happened to the CI, but it may be related to my recent changes.

Gadgetoid:
Looks like the storm has calmed now. Not seen it broken quite like that before!

Thank you again, saved my bacon here.

PR #781, opened 2023-05-18
CI: Add workflow version to cache key.

Gadgetoid:
This formalises a sort of cache-buster we can bump if the workflow changes significantly. This ensures the cache key remains consistent between the dependency and individual jobs so we don't get a race condition where a dirty cache will break subsequent builds.

The hacks to revert Pico SDK / MicroPython to its vanilla state have been removed, too, since this *should* be guaranteed by the (hopefully now clean) cache.

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   MICROPYTHON_VERSION: v1.20.0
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Rolling the dice on this, the only way we're going to find out if it works is by giving it a shot.

PR #790, opened 2023-06-19
Fix out of box failure when ARM toolchain is not in path

awjlogan:
When the ARM toolchain is not on the path and is specified with the environment variable `PICO_TOOLCHAIN_PATH`, linking in the pico_explorer example fails as it can't find `arm-none-eabi-ld`.

  - Add expansion of PICO_TOOLCHAIN_PATH for arm-none-eabi-ld into base `CMakeLists.txt`

Tested on Linux (Kubuntu 22.04) and macOS (Monterey).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ function(add_resource target file)
             ${CMAKE_CURRENT_SOURCE_DIR}/${FILENAME}
             ${CMAKE_CURRENT_BINARY_DIR}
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you. I have no earthly idea why the CI has given up the ghost here.

I think the canonical solution to this is to switch to Pico SDK's compiler search function:

```cmake
pico_find_compiler(PICO_COMPILER_CC ${PICO_GCC_TRIPLE}-ld)
```

So we don't have to worry about handling edge cases or weirdness. I've opened #815 with this change.

PR #791, opened 2023-06-22
Document PicoGraphics fixed_width

### LINKED ISSUE #765: FR: Add a monospace bitmap font to PicoGraphics
Looking at their definitions in https://github.com/pimoroni/pimoroni-pico/tree/cb5db1d5944d0b6adc9ef9b7c014749b03301deb/libraries/bitmap_fonts , all the available fonts for PicoGraphics---bitmap6, 8, and 14_outline---are proportionally spaced (variable width).

I'd like a monospace font, so that debug readouts and the like in Tufty2040 scripts that have rapidly-updating numbers don't jump about (since "1" is narrower than other digits). A monospace variant of bitmap8 would be just fine, for example.

Being able to provide the font data via the Python API would natch solve this, and "can I have another font please" in the general case. It looks...plausible? to implement that, with perhaps some awkard corners with lifespan/deallocation. (But I don't have the build env to try.)

LionsPhil:
(VSCode also stubbornly fixed some trailing space.)

Closes #765.

### DIFF
--- a/micropython/modules/picographics/README.md
+++ b/micropython/modules/picographics/README.md
@@ -305,7 +305,7 @@ These are aligned from their top-left corner.
 * `bitmap8`
 * `bitmap14_outline`
 
-Vector (Hershey) fonts. 
+Vector (Hershey) fonts.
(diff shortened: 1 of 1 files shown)


helgibbons:
Lovely, thank you!

PR #793, opened 2023-06-23
Enviro Pico description generation did not match f-string rounding

awjlogan:
Description generation (eg, humidity -> "good" or "bad") did not match the f-string formatting, which rounds to the nearest value.


Gadgetoid:
Good catch, thank you!

PR #797, opened 2023-07-04
Make Galactic clock example responsive

North101:
When pressing a button on the Galactic clock example (e.g. changing brightness, changing utc offset), it will often take a bit to respond. This is because the redraw code checks whether the second has changed.

This PR just unsets `last_second` when certain buttons are pressed allowing the display to update quicker when required


Gadgetoid:
Good catch, thank you!

PR #807, opened 2023-07-31
 Added light sensor example with auto brightness feature

makosa-irvin:
Uses the onboard sensor on the unicorn to detect the brightness
Displays the percentage of the brightness value obtained


helgibbons:
Lovely, thank you!

PR #809, opened 2023-08-02
Added light sensor example to the cosmic and stellar unicorns

makosa-irvin:


helgibbons:
Looks great, thank you. I like the ☀!

Few very minor things I noticed:

- The hue/brightness adjustment seems quite flickery at certain brightnesses (I think the LED lights in this room are particularly unforgiving!) You might be able to do some averaging on your brightness read to smooth out the values to be more visually appealing? I think [this Tufty example](https://github.com/pimoroni/pimoroni-pico/blob/main/micropython/examples/tufty2040/autobright.py) does this by taking into account the last brightness reading.
- This makes for a more symmetrical % symbol:
```# draws percentage icon
def draw_percentage(x, y):
    graphics.rectangle(x + 1, y + 1, 2, 2)
    graphics.line(x + 1, y + 5, x + 6, y)
    graphics.rectangle(x + 4, y + 4, 2, 2)
```
- On the Cosmic example, it might be nice if the height of the brightness bar was the same as the height of the sun icon. I'd also nudge the icon and bar up a row of pixels, so they're equally spaced between the two rows of text.

ZodiusInfuser:
Thank you for this!

PR #813, opened 2023-08-10
fix: Only set time if the wlan is connected

andrewjw:
Previously, if we dropped out of the wlan loop early because of an error (wlan.status() < 0) it would still print "Connected", and try to set the time.


Gadgetoid:
Thank you!

Issue #826, opened 2023-08-28
CMake files in (some?) libraries don't pick up include directories (e.g. servo2040)


ahnlak:
(from a question on Discord, but the deeper I dig the more confused I get!)

So, the Pico Boilerplate should allow us to pull in `pimoroni-pico` libraries simply by adding (for example):

```
include(libraries/servo2040/servo2040)
target_link_libraries(MyAwesomeApp servo2040)
```

BUT, if that library uses other libraries or drivers the provided `servo2040.cmake` does not expose the include directories of those sub-requirements.

While you can work around this by manually `include`ing all the drivers in your application's CMakeLists.txt that feels wrong (and involves a lot of digging around to find out what's required).

Some libraries (for example, `pico_explorer`) have suitable `include` directives to pick up the drivers that are required - my instinct is that this just got missed out in more recent libraries and, because the problem only really comes up when you're doing out-of-tree builds it probably doesn't upset the automated builds and until someone tries to use them in anger, we're none the wiser.

(I can't help feeling that there might be a more "CMake-y" way of doing it, but I haven't found it yet)

TL:DR - add `include` directories to library / driver cmake files to ensure they find required include files in out-of-tree builds.

I'm happy to go through and apply such a fix to places I can find, but as it involves touching a lot moving parts I thought I'd raise an issue first in case I've missed something obvious!

Gadgetoid:
I distinctly remember previous a PR or Issue related to this somewhere, but I can't seem to dig it up.

At the time I think I hand-waived it away as needing some kind of "pimoroni_pico_init()" to set up all the paths so that everything would just work, but I think you make a good point- that some drivers set up include paths while others do not. Would make more sense - and be less complicated - to just fix this and make things consistent.

PR #829, opened 2023-08-31
Added includes to allow for libraries to be linked to out-of-tree

ahnlak:
As per https://github.com/pimoroni/pimoroni-pico/issues/826; this PR adds appropriate `include` statements into all the library `cmake` files (that I could find that required them), so that out-of-tree builds (such as those based on the boilerplate) only need the libraries themselves added to the project's `CMakeLists.txt`

Previously, to successfully compile against a library, you had to manually add those includes to your own project's `CMakeLists.txt` for all dependent libraries and drivers, which was an exercise in exploration and a bit of a barrier for anyone less confident with hacking cmake around.

(the `include` statements are deliberately placed at the top of the file, because otherwise you get into a mess because a lot of the cmake files re-use the same variable names)

This (necessarily) touches a lot of files, but only slightly so hopefully it's not as big a PR as it looks!


Gadgetoid:
I think the CI CMake build is pretty much the litmus test for if anything serious has gone awry, and it seems to be working. Quite painful that they all need a guard against defining the target twice, but that makes sense to avoid the existing workarounds breaking.

Didn't expect a PR for this, but warmly welcome it, thank you!

PR #830, opened 2023-09-04
Added some PicoGraphics based MP examples for the PicoUnicorn

ahnlak:
As mentioned on the [forum](https://forums.pimoroni.com/t/pico-unicorn-pico-graphics/22811/4), the examples for the Pico Unicorn Pack used the 'native' graphics, rather than the newer PicoGraphics based ones.

Duplicated the two existing examples into `-picographics` versions, and added a version of `fire-effect.py` from the other Unicorns' examples, turned on it's side to get a bit more height displayed!

Also updated the PicoGraphics readme to (a) acknowledge the existence of the Pico Unicorn, and (b) mention that the special `update()` handling applies to *all* Unicorns, not just the Galactic.

### DIFF
--- a/micropython/examples/pico_unicorn/demo-picographics.py
+++ b/micropython/examples/pico_unicorn/demo-picographics.py
@@ -1,0 +1,38 @@
+# This is a PicoGraphics version of the original demo.py
+
+from picounicorn import PicoUnicorn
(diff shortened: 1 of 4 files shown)


helgibbons:
That would be super, thanks!

PR #838, opened 2023-09-08
Fix clean Micropython build

MichaelBell:
I had trouble with my local MicroPython build for ages until @Daft-Freak suggested wrapping the `pio.h` includes like this..  I feel this should be in the standard build to avoid others having the same pain!

### DIFF
--- a/drivers/encoder/encoder.cpp
+++ b/drivers/encoder/encoder.cpp
@@ -4,7 +4,10 @@
 #include "hardware/irq.h"
 #include "hardware/clocks.h"
 #include "encoder.hpp"
+
+#ifndef NO_QSTR
 #include "encoder.pio.h"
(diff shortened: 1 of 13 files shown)


Daft-Freak:
Hah, I'd entirely forgot about this.

Think this works around the fact that the QSTR preprocessing doesn't inherit the dependencies of the main build... by taking advantage of the code not needing to be entirely valid while that's happening. Should be fine unless there's a QSTR in a generated header for some reason.

(Then again, not a python person so possibly speaking complete nonsense :shrug:)

Gadgetoid:
Ah I've been just charging on through this every single time :laughing:  thank you. I'm glad it finally annoyed someone enough to stop, step back and fix it.

PR #839, opened 2023-09-08
Allow pen function for fast tile alpha blending

MichaelBell:
This is just a hook to allow the pen implementation to provide a fast alpha blend for a picovector tile.

### DIFF
--- a/libraries/pico_graphics/pico_graphics.hpp
+++ b/libraries/pico_graphics/pico_graphics.hpp
@@ -293,6 +293,8 @@ namespace pimoroni {
     virtual void frame_convert(PenType type, conversion_callback_func callback);
     virtual void sprite(void* data, const Point &sprite, const Point &dest, const int scale, const int transparent);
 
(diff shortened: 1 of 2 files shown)


libraries/pico_graphics/pico_graphics.hpp
```diff
@@ -293,6 +293,8 @@ namespace pimoroni {
     virtual void frame_convert(PenType type, conversion_callback_func callback);
     virtual void sprite(void* data, const Point &sprite, const Point &dest, const int scale, const int transparent);
 
+    virtual bool render_pico_vector_tile(const Rect &bounds, uint8_t* alpha_data, uint32_t stride, uint8_t alpha_type) { return false; }
```
MichaelBell, line 296: Not sure if it would be preferable to expose the pretty_poly antialias type here instead of having magic numbers that match that enum?

Gadgetoid: Have to be careful about too tightly coupling PicoGraphics and PicoVector since it's entirely possible to have C++ or MicroPython builds that include the former but not the latter. In this case perhaps it's PicoGraphics that should define the enum type for anti-aliasing so that PicoVector can inherit it.

PR #841, opened 2023-09-10
Initialize all matrix values

MichaelBell:


Gadgetoid:
I will just have a quiet cry at how many times I skipped past this line of code now. Thank you for spotting it!

PR #842, opened 2023-09-10
Fix newlines in rotated text

MichaelBell:
Carriage return went upwards!


Gadgetoid:
This will absolutely teach me not to test with:

```python
"hello world\nhello world"
```

Thank you!

PR #850, opened 2023-09-21
IOExpander: add I2C address change example

helgibbons:


Gadgetoid:
It took me an embarrassingly long time to realise the address change lines were commented out...

PR #860, opened 2023-10-09
Badger2040/2040W: Remove old/incompatible examples.

Gadgetoid:
Get rid of broken/old examples to prevent confusion.

Do any of these need porting to the new Badger 2040?

Badger now lives at: https://github.com/pimoroni/badger2040


Gadgetoid:
Should probably update the README.md to include a deprecation message and a link to the Badger repository!

PR #867, opened 2023-10-26
Add support for 96x48 display to Interstate75

ConnorLinfoot:
Adds support for the slightly more unconventional display type of 96x48 for Interstate75 `DISPLAY_INTERSTATE75_96X48`. Personally using this with the matrix from Waveshare: https://www.waveshare.com/wiki/RGB-Matrix-P2.5-96x48-F


Gadgetoid:
Thank you!

PR #873, opened 2023-11-07
Expose set_pixel to MicroPython

tmsmr:
While using Pico Graphics might usually be the proper way to draw on the Stellar/Galactic/Cosmic Unicorn, there are use cases where setting the pixels directly is desirable.
This PR exposes `GalacticUnicorn::set_pixel` (https://github.com/pimoroni/pimoroni-pico/blob/v1.21.0/libraries/galactic_unicorn/galactic_unicorn.cpp#L447) to MicroPython.

The implementation is pretty much copy-pasted from https://github.com/pimoroni/pimoroni-pico/blob/v1.21.0/micropython/modules/pico_unicorn/pico_unicorn.cpp#L56.

If this change generally suits you, i'll add it to the Stellar and Cosmic Unicorn as well and update the Docs accordingly as preparation for the merge.

### DIFF
--- a/micropython/modules/galactic_unicorn/galactic_unicorn.c
+++ b/micropython/modules/galactic_unicorn/galactic_unicorn.c
@@ -20,6 +20,7 @@ MP_DEFINE_CONST_FUN_OBJ_KW(Channel_play_tone_obj, 2, Channel_play_tone);
 
 MP_DEFINE_CONST_FUN_OBJ_1(GalacticUnicorn___del___obj, GalacticUnicorn___del__);
 MP_DEFINE_CONST_FUN_OBJ_1(GalacticUnicorn_clear_obj, GalacticUnicorn_clear);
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Sorry for the delay feeding back on this. This seems reasonable, though there may be some confusion that arises between the two methods of drawing to the display. Do you have any use-cases or examples?

If you're still interested do you want to add the other two displays to this PR?

tmsmr:
The use-case is for my personal project to be honest: https://github.com/tmsmr/RUC

This is basically a minimal TCP-based protocol and a client lib which allows to draw in a "frame-buffer" on the client side and transmit only the diff required to update the image.

Every `update`-call in uPython on the Pico overwrites the complete frame-buffer: https://github.com/pimoroni/pimoroni-pico/blob/main/libraries/galactic_unicorn/galactic_unicorn.cpp#L517

While this might not actually be a performance issue, it feels unnecessary for me in cases where only some pixels have changed...

If the change generally suits you, i'll add the other Displays as well and update the docs accordingly. I could add a note, that using `update` should be preferred generally.

PR #877, opened 2023-11-15
add adjustment for time zone offset

pacohope:
This adds a really primitive adjustment for time zones. The user can simply set the value for `tz_offset` to the hours difference to UTC and it will do a really simple calculation. This isn't real time zone support, but it lets a new user at least see this demo in their current time zone.


Gadgetoid:
Thank you!

PR #878, opened 2023-11-16
CI: Move some workflow steps into ci/micropython.sh

Gadgetoid:
For a few reasons:

1. Reduce our dependency on GitHub actions workflows
2. Attempt to provide tools we can run locally to replicate a CI build
3. Make build steps more discoverable and less intertwined with GitHub Actions idiosyncrasies

This changeset also switches to `arm-none-eabi-gcc-action` (instead of apt) to handle installing the ARM GCC toolchain. This uses caching, speeding up the compiler install from 60+ seconds down to about 7 seconds.

I have dropped our "deps" step (a pre-requisite of each MicroPython build) in favour of just setting up the environment for each build. This is - in theory - slower, but is much less complicated and less prone to weird errors.

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -59,13 +59,18 @@ jobs:
         restore-keys: |
           ccache-micropython-${{ matrix.name }}-${{ github.ref }}
           ccache-micropython-${{ matrix.name }}-
-  
+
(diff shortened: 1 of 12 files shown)


Gadgetoid:
Rebased on top of #892

PR #888, opened 2023-12-29
[Inky Frame] Fix news headline redirect example

everyplace:
This very small PR accomplishes two things:

1. Swaps the protocol from `http` to `https` for BBC's example rss feeds. The current feed requests were failing because BBC has implemented a redirect on RSS to enforce `https`.
2. Adds a comment for the 7.3 frame size, to match the existing ones for the other Inky Frame sizes.


helgibbons:
Super, thanks very much for this :)

Issue #889, opened 2024-01-01
I2C pins are undocumented in motor2040 header file


robberwick:
The `motor2040.hpp` file declares variables for all the connected pins on the board, but omits definitions for the pins connected to the SDA/SCL pins of the QW/ST connector (GPIO20 & GPIO21 respectively). These values can be found by looking at the schematics, but It would be really useful to add these so that it's possible to access them via the `motor::MOTOR2040` namespace.

ZodiusInfuser:
Hi @robberwick,

Thank you for raising this. The reason these pins were not defined in the `motor::MOTOR2040` namespace is because they are accessible through the `I2C` class within `pimoroni_i2c.hpp`. This is our convenience class for dealing with I2C comms, which simplifies some of the operations involved in reading and writing registers.

One of its constructors accepts `BOARD` enum value from `pimoroni_common.hpp` of which MOTOR_2040 is one of those already defined. Does this work for you? If not I will have a look at your PR.

robberwick:
Hi @ZodiusInfuser 

Ok, if that's the convention, then I guess it's worth sticking to that. I would just say that as a user of that board, I wasn't aware of that mechanism, so I would suggest that it probably needs to be more discoverable. For my money, I would still define the pins in that header file, since that seems to be the most obvious place to go looking for them (at least it seemed that way to me), but perhaps also add a comment on there to point the user at the i2c helper + board definition documentation so that they can choose which is most appropriate for them.

If that's how you decide to go, and you want me to try to write a suitable comment, I can update my PR accordingly.

ZodiusInfuser:
Closing this as the associated PR is now merged.

PR #890, opened 2024-01-01
Add I2C pin definitions to motor2040 header

robberwick:
Add definitions for the QW/ST connector SDA & SCL pins

addresses #889

### DIFF
--- a/libraries/motor2040/motor2040.hpp
+++ b/libraries/motor2040/motor2040.hpp
@@ -47,10 +47,11 @@ namespace motor {
     const uint LED_DATA = 18;
     const uint NUM_LEDS = 1;
 
-    const uint USER_SW = 23;
-
+    const uint I2C_INT = 19;
     const uint I2C_SDA = 20;
     const uint I2C_SCL = 21;
+
+    const uint USER_SW = 23;
 
     const uint ADC_ADDR_0 = 22;
     const uint ADC_ADDR_1 = 24;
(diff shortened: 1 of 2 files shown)


ZodiusInfuser:
Thanks for this. As mentioned in issue #889 the intention was for people to access these pins via our I2C class, which has many convenience functions for dealing with I2C comms. I realise though that there is no harm in having these pins defined in `motor::MOTOR2040` for those wanting to roll their own solution though.

Before I approve this PR, could you amend the following comments.

Also, could you include the same three lines for SERVO2040 too please?

libraries/motor2040/motor2040.hpp
```diff
@@ -49,6 +49,9 @@ namespace motor {
 
     const uint USER_SW = 23;
 
+    const uint I2C_SDA = 20;
```
ZodiusInfuser, line 52: Can these lines be put before USER_SW, so they are in number order?
Also `const uint I2C_INT = 19;` should be included too

robberwick: `I2C_INT` added and reordered in  [3e81b24](https://github.com/pimoroni/pimoroni-pico/pull/890/commits/3e81b245a1a16e3d791bf92210b822b52a7791ef)
`i2c_*` pin defs added to `servo2040.hpp` in [9e6a072](https://github.com/pimoroni/pimoroni-pico/pull/890/commits/9e6a0725c03daaaa8777a4e798c8c13fabf61000)

robberwick:
@ZodiusInfuser Just checking this is ok, and there's nothing else you want me to do before it's mergeable?

ZodiusInfuser:
Sorry for the delay. Yes, this is good. Thanks for the contribution!

PR #897, opened 2024-01-19
Experimental: builds with dual FAT MSC support.

Gadgetoid:
Highly experimental mass-storage supporting builds of Pico firmwares. Uses a dual device setup with one readable to the host and one readable to the device.

:warning: Don't flash these builds on your Pico unless you've backed up and run `flash_nuke.uf2`! They will break your filesystem and do terrible awful things.


alphanumeric007:
Kind of like how Circuit Python does it? I can see this ability as a bonus on devices like the Pico Lipo 16mb, or the Tiny 8mb.

Gadgetoid:
> Kind of like how Circuit Python does it?

Yes, but with none of the careful consideration the CircuitPython team put into making it work at least somewhat well, and documenting how to use it :laughing: 

Though I don't think they make a habit of splitting the filesystem like this- it's always readable to the host and write-locked for the device by default, IIRC, and can be changed from Python if you need to write log files or something?

PR #899, opened 2024-01-23
JPEGDEC: Backport width/height changes from pngdec.

Gadgetoid:
Open JPEG file or stream to read width/height before decode.

Replaces #895


Skyler84:
Have got this running on my inky frame now, works very nicely to allow centering of images! Will close #895 in anticipation of this being merged. Thanks!!

Issue #866, opened 2023-10-25
Update of graphics.update() seems to get stuck indefinatelly


klausi83:
I have written a py-script that should convert inky_frame to a small self contained picture frame. (Inky_frame 7.3). Unfortunately I get an unexplained hang-up.
Code: https://github.com/klausi83/inky_frame
The code seems to run ok, for one or two pictures after first initialization (coming from main.py and selecting Button "C"). The event inky_frame.woken_by_rtc() triggers but almost always, the update of the displayed picture gets stuck on line 124 and 127 (most likely in line 127). A press of the "reset" usually does not help, it gets stuck at the same point.

Getting out of this status is: press reset + (A and E) and start from main.py again.

Is there a way to debug this deeper, or does someone know a solution?

klausi83:
I tested some more today. Tried twofold
1. Increasing update frequency to 2 Minutes. Running it via USB and Thonny for more than an hour. All graphics-updates worked like they are supposed to.
2. Increased update frequency to 2 Minutes. Running without USB connection on a variable power supply. Tried everything between 2,8V and 5,4V. In all cases the controller and the display worked. Running the code yielded the above error every trial run after only a few graphics updates.
So, to me, there seems to be some mayor difference in graphics.update() routine depending on inky-frame running in USB-Mode or running in stand-alone mode. Is someone more knowledgeable than me and can hint me in a direction to further analyze or rectify this?

alexmuller:
I am also having this problem with an Inky 7.3

I have a very small `main.py`:

```
import time

from picographics import PicoGraphics, DISPLAY_INKY_FRAME_7 as DISPLAY

# A short delay to give USB chance to initialise
time.sleep(0.5)

# Setup for the display.
graphics = PicoGraphics(DISPLAY)
WIDTH, HEIGHT = graphics.get_bounds()
graphics.set_font("sans")

graphics.set_pen(1)
graphics.clear()
graphics.set_pen(0)

graphics.set_thickness(2)

graphics.text('Hello', 20, 40, 600, 2)
graphics.text("Today this thing is happening", 20, 90, 600, 1)

graphics.text('Shopping list', 20, 160, 600, 2)
graphics.text('Thing one', 20, 210, 600, 1)
graphics.text('Thing two', 20, 240, 600, 1)

print('Updating graphics')
graphics.update()
print('Done')
```

In the Thonny output I see:

```
MPY: soft reboot
Updating graphics
```

[...]

PR #900, opened 2024-01-31
Circumvent hangs on update method of inky frame 7 when on battery power

w3stbam:
Some users and myself experienced issues with the update method, which often hangs when the inky frame is powered by batteries.
Discussions:
https://github.com/pimoroni/pimoroni-pico/issues/866
https://forums.pimoroni.com/t/inky-frame-not-refreshing-screen-on-battery-power-using-inky-frame-sleep-for/23174/23

This adds a timeout for the busy_wait function, so that the code continues after a default timeout of 45 seconds. It seems it fixes the issues, as there are no more hangs. Of course, the root cause seems to be somewhere else, but this way we can at least use the update method successfully on battery power.

Relases for testing pruposes can be found here:
https://github.com/w3stbam/pimoroni-pico/releases/

### DIFF
--- a/drivers/inky73/inky73.cpp
+++ b/drivers/inky73/inky73.cpp
@@ -43,11 +43,20 @@ namespace pimoroni {
     TSSET = 0xE6 // E5 or E6
   };
 
+  int timeout = 0;
+
   bool Inky73::is_busy() {
-    return !(sr.read() & 128);
+    bool busy = !(sr.read() & 128);
+    if (!busy) {
+      return false;
+    } else if(absolute_time_diff_us(get_absolute_time(), timeout) < 0) {
+      return false;
+    }
+    return true;
   }
   
-  void Inky73::busy_wait() {
+  void Inky73::busy_wait(uint maximum_wait_ms=30000) {
+    timeout = make_timeout_time_ms(maximum_wait_ms);
     while(is_busy()) {
       tight_loop_contents();
     }
@@ -56,7 +65,7 @@ namespace pimoroni {
   void Inky73::reset() {
     gpio_put(RESET, 0); sleep_ms(10);
     gpio_put(RESET, 1); sleep_ms(10);
-    busy_wait();
+    busy_wait(5000);
   }
    
   void Inky73::init() {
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Thank you! And sorry for missing this before- I'd been embroiled in Pi 5 chaos.

The public API for `busy_wait` should probably not change, so the default value - `=45000` - for `maximum_wait_ms` should probably be in the function declaration (.hpp) and not the definition.

You could probably also leave `is_busy()` as-is, and move the timeout handling to `busy_wait`, since `is_busy()` is in the public API so it could be called without setting a `timeout` value and, I guess, would have unexpected results.

`make_timeout_time_ms` also returns a `absolute_time_t ` (alias of `uint64_t` afaik) rather than an `int`, I think there would be a narrowing conversion and a signedness issue here so something like:

```c++ 
  // hpp
  void busy_wait(uint maximum_wait_ms=45000);
  // cpp
  void Inky73::busy_wait(uint timeout_ms) {
    absolute_time_t  timeout = make_timeout_time_ms(timeout_ms);
    while(is_busy() && !time_reached(timeout)) {
      tight_loop_contents();
    }
  }
```

Note: `time_reached()` looks handy, TIL- https://www.raspberrypi.com/documentation//pico-sdk/hardware.html#ga1e41eca96d5c1b3b5e69e3b1a00cdf7d

These e-ink displays are awkward, I have a timeout-on-busy-wait pattern in most of the Python-equivalent code, so it makes a lot of sense here too. Thank you.

stuartleeks:
Thank you for this. I've been having issues with my Inkyframe hanging on updates but it has been working perfectly since applying the firmware in from your test releases!

I look forward to this being in a official release 😃

Gadgetoid:
Replaced by #900 implementing my suggestions.

I was going to add a `Co-authored-by` but I don't think `f <you@example.com>` is your real email address 😆  so I settled for attribution via the commit message.

Did get a chuckle out of it though.

kevinjwalters:
@Gadgetoid I'm not seeing the mentions on the [release page](https://github.com/pimoroni/pimoroni-pico/releases). What version is this in?

This fix appears to be needed for a 7.3" Inky Frame based on https://forums.pimoroni.com/t/inky-frame-slide-show/19996/35

PR #907, opened 2024-02-27
Fixes for PNGDEC on Badger 2040 / Badger 2040 W

Gadgetoid:
These fixes are required for PNGDEC to function correctly on Badger 2040 and Badger 2040 W, so it can be added - https://github.com/pimoroni/badger2040/pull/75

The fix to `PicoGraphics_Pen1BitY::set_pen` corrects handling of RGB and ARGB PNGs.

The fix to PNGDEC adds explicit support for 1bit PNGs, which would obviously be desirable on a 1bit display.

### DIFF
--- a/libraries/pico_graphics/pico_graphics_pen_1bitY.cpp
+++ b/libraries/pico_graphics/pico_graphics_pen_1bitY.cpp
@@ -15,7 +15,7 @@ namespace pimoroni {
   }
 
   void PicoGraphics_Pen1BitY::set_pen(uint8_t r, uint8_t g, uint8_t b) {
-    color = std::max(r, std::max(g, b));
+    color = std::max(r, std::max(g, b)) >> 4;
   }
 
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I've run through some tests rendering PNGs on Badger 2040 W with these changes. Tried what I hope are 1bit, 2bit, 4bit and 8bit indexed plus RGB with the desired results.

Ignore the...weirdness of these test images, I missed the "include unused colours" tickbox in GiMP's colour conversion dialog and uh, got creative....

1bpp
badge-1bpp

2bpp
badge-2bpp

4bpp
badge-4bpp

8bpp
badge-8bpp

rgb
badge-rgb

rgb
badgerpunk

Note: In all cases these will be crushed to 1bit since Badger 2040 is monochrome only, but RGB or higher colour depths should use the 16 step dithering patterns. The most obvious way to see this in effect is the RGB badgerpunk full screen image.

Issue #576, opened 2022-11-13
Rounding of brightness values on Galactic Unicorn micropython class


Footleg:
Playing with the micropython examples for Galactic unicorn I noticed that brightness values show rounding errors.
```
gu.set_brightness(0.5)
print(f"Brightness set to {gu.get_brightness()}")
```
returns:
Brightness set to 0.5019608

Gadgetoid:
Looks like it's multiplied up to an 8-bit by 256 and divided down by 255 :grimacing: 

https://github.com/pimoroni/pimoroni-pico/blob/main/libraries/galactic_unicorn/galactic_unicorn.cpp#L502-L510

Gadgetoid:
Should hopefully be fixed by #911, floating point quantisation issues notwithstanding.

SOMEBODY CALL NINE ONE ONE! Shawty fire burning on the dance floor, whoa!

PR #911, opened 2024-03-11
G/S/C Unicorn: Fix get_brightness to use correct max value.

Gadgetoid:
Add a comment noting that 256 is the correct maximum brightness.


Gadgetoid:
This probably wont fix the floating point rounding issue, but since the maximum brightness is 256 (clarified with a comment) then returning that value divided by 255 in `get_brightness()` would be incorrect.

Should "fix" at least part of https://github.com/pimoroni/pimoroni-pico/issues/576

PR #917, opened 2024-03-26
PNGdec: Sync with upstream.

Gadgetoid:
* Fixed pixel conversion of 1-bpp grayscale to RGB565
* Improved perf for systems which allow unaligned memory accesses
* corrected optimized code to not go past buffer end
* speed improvements


Gadgetoid:
Okay I've done some prodding and poking, but I can't figure out why this stalls at decoding. Since the current PNG implementation is working well for us, this is a problem for Future Phil.

PR #927, opened 2024-04-08
CI: Better build size debugging

Gadgetoid:
This is a little bit of a redux of the issue I raised here: https://github.com/micropython/micropython/issues/8680

And the PR here: https://github.com/micropython/micropython/pull/8761

Neither of these have had any traction for a while, due to an impasse with the way flash sizes are declared with Pico's `bi_decl` being incompatible with how we might want to supply (or retrieve from) flash sizes in `memmap_mp.ld`.

The general idea is to split `FLASH` into `APP` and `FILESYSTEM` so we have an early failure if the two happen to overlap.

Without a linker error to catch this, a MicroPython build will generate a filesystem at runtime and overwrite itself in potentially catastrophic ways.

Making this tooling work requires adding an `memmap_mp.ld` to each board directory. Since this does not inherit the flash size from `mpconfigboard.h` it removes the previous "single source of truth" for flash size and... this can easily catch you out.

In all cases the linker variable `_flash_app_size` should be `ACTUAL_FLASH_SIZE - MICROPY_HW_FLASH_STORAGE_BYTES`. Eg: For Pico this is `2048 (2MB) - 1408 (1.4MB) = 640k`.

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -98,6 +98,12 @@ jobs:
         source $BUILD_TOOLS
         hack_patch_micropython_print_memory_usage
 
+    - name: "HACK: CMakeLists.txt Enable Custom Linker Patch"
+      shell: bash
+      run: |
+        source $BUILD_TOOLS
+        hack_patch_micropython_enable_custom_linker
+
     - name: "HACK: Pico SDK Patch"
       shell: bash
       run: |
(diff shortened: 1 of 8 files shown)


Gadgetoid:
Judging by the before and after of Tiny 2040 (8MB) the ulab module - as configured - weighs in at about 92k.

Gadgetoid:
Having a long, hard think about how much appetite I have to babysit Yet Another Patch 😆

PR #933, opened 2024-04-17
CI: Scannable artifact and release names.

Gadgetoid:
When I go to fetch an artifact for a MicroPython build my brain is always confounded by the repetition of "pimoroni-" before each file.

Change the names so the product is up-front, so the list is easier to scan-read through.

This will break: https://github.com/pimoroni/pimoroni-pico/blob/main/setting-up-micropython.md (TODO: fix in this PR)

And probably: https://learn.pimoroni.com/article/getting-started-with-interstate-75

And: https://github.com/pimoroni/enviro/blob/b5bff82f9001409c9683be43b9f2114e7da8b430/.github/workflows/release-zip.yml#L16

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -42,7 +42,7 @@ jobs:
 
     env:
       # MicroPython version will be contained in github.event.release.tag_name for releases
-      RELEASE_FILE: pimoroni-${{ matrix.name }}-${{ github.event.release.tag_name || github.sha }}-micropython
+      RELEASE_FILE: ${{ matrix.name }}-${{ github.event.release.tag_name || github.sha }}-pimoroni-micropython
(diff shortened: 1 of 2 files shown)


Gadgetoid:
@helgibbons gentle nudge that me merging this might break some learning portal articles, since the product names change from "pimoroni-product-version-micropython.uf2" to "product-version-pimoroni-micropython.uf2".

See: https://github.com/pimoroni/pimoroni-pico/actions/runs/8721551227?pr=933

helgibbons:
Ta - I've updated a few things on Learn to allow for this change.

Think the file names on [this page](https://github.com/pimoroni/pimoroni-pico/blob/main/setting-up-micropython.md) might still not be quite right - looks like some of them have rogue 'picow's in front of them 🐄

PR #942, opened 2024-05-30
Patch inky launcher

thirdr:
- Applied the fixes from [here](https://github.com/pimoroni/picovision/pull/97) to news headline example.
- `get_rss()` now returns an empty list on failure rather than False.


Gadgetoid:
Thank you! :partying_face:

Issue #504, opened 2022-08-30
BME280 example displays ER when result is read fine


johnvanham:
Hello, I think this looks like a bug? Or have I misunderstood.

The example for the BME280 loads a reading from the board in forced mode but then the output checks whether results.status == BME280_OK

BME280_OK is set in the driver as 0

Yet status is a boolean set by the read_forced() function. So it prints ERR because 0 = false when actually it is OK?

https://github.com/pimoroni/pimoroni-pico/blob/05ef0d7f037ad5050d8612d23c9829e0e6f29204/examples/breakout_bme280/bme280_basic.cpp#L25-L26

https://github.com/pimoroni/pimoroni-pico/blob/05ef0d7f037ad5050d8612d23c9829e0e6f29204/drivers/bme280/bme280.cpp#L59

thirdr:
This has now been fixed in #943

PR #943, opened 2024-05-31
Fixed result status bug

thirdr:
Fixed the bug in the BME280 examples reported in #504


Gadgetoid:
Thank you! These bugs wont know what hit 'em!

Issue #505, opened 2022-09-01
Tufty2040 C++ drawing example buttons created with wrong polarity


Footleg:
I found the buttons on the Tufty2040 badge did not work in my code based on the drawing example code. I tried copying from other examples which used the buttons on other boards with

```
    if(button_a.read()) {
        //Do stuff
    }
```

But neither this, or trying to read button_a.raw() appeared to ever return true. A hint from Phil in a Twitter reply enabled me to work out how to change the button definitions in the example so that they work. I've tested with all 5 buttons. If you define the buttons as follows with the Polarity High then the button.read() call returns true when the button is pressed:

```
Button button_a(Tufty2040::A,Polarity::ACTIVE_HIGH);
Button button_b(Tufty2040::B,Polarity::ACTIVE_HIGH);
Button button_c(Tufty2040::C,Polarity::ACTIVE_HIGH);
Button button_up(Tufty2040::UP,Polarity::ACTIVE_HIGH);
Button button_down(Tufty2040::DOWN,Polarity::ACTIVE_HIGH);
```

Suggesting this is updated in the example for tutfy2040

bengriepp:
Can this please be added, took me so long to figure this out

PR #944, opened 2024-06-03
Fixed button polarity

thirdr:
Updated button example code so the polarity is correct and returns true when the buttons are pressed. As raised in #505


Gadgetoid:
Thank you!

Issue #618, opened 2023-01-04
MICS6814 c implementation has the OX and RED pins defined the wrong way round


knightinlet:
Looks like the c implementation has the OX and RED pins defined the wrong way round.

breakout_mics6814.hpp has the values 

    static const uint8_t MICS_VREF        = 14;
    static const uint8_t MICS_RED         = 12;
    static const uint8_t MICS_NH3         = 11;
    static const uint8_t MICS_OX          = 13;

whereas the python implementation 

pimoroni/mics6814-python/library/mics6814/__init__.py

MICS6814_VREF = 14  # P1.4 AIN0 - 3v3
MICS6814_RED = 13   # P0.7 AIN2 - Reducing
MICS6814_NH3 = 11   # P0.6 AIN3 - NH3
MICS6814_OX = 12    # P0.5 AIN4 - Oxidizing

Looking at the component schematic it looks like the python implementation is correct.

PR #945, opened 2024-06-03
Corrected pin definitions for OX and RED

thirdr:
As raised in #618.

Confirmed correct pins as shown in the schematic [here](https://cdn.shopify.com/s/files/1/0174/1800/files/MICS6814_breakout_schematic.pdf?v=1628772412)


Gadgetoid:
Thank you!

PR #956, opened 2024-06-12
Added support for 16x16 LED Matrix

mwilliams03:
Updated to support 16x16 RGB LED matrix


Gadgetoid:
Which 16x16 LED matrix? 🤔

mwilliams03:
A new board from OzzMaker.
We have yet to add it to our website, but here are two images.
https://ozzmaker.com/wp-content/uploads/2024/06/IMG_5849.jpg
https://ozzmaker.com/wp-content/uploads/2024/06/IMG_5848.jpg

Should be listed on our site by the end of next week.

Gadgetoid:
Oooh is that a flat flax interconnect? Interesting!

I really need to do something about how inflexible the i76 code is. Making wide marquee displays from these panels would be neat!

PR #957, opened 2024-06-12
add brightness control to default Plasma Stick W example

helgibbons:
Makes the default Plasma Stick example a little more conservative with its brightness. This should hopefully stop it from drawing a load of current and making your USB controller fall over if you have an older computer (or are programming it using a Pi :) )


Gadgetoid:
Ooof! Thank you

PR #960, opened 2024-06-14
Add SPI import to SD card upload example

johnbintz:
As the code exists now, you won't be able to mount the SD card due to the missing import.


helgibbons:
Lovely, thank you for that!

Issue #965, opened 2024-07-02
ADC(29) needs to be ADC(Pin.29) in recent versions of MicroPython


helgibbons:
As per: https://forums.pimoroni.com/t/pico-lipo-not-reading-battery-voltage/25060/

More info at https://github.com/micropython/micropython/issues/10947

Gadgetoid:
I think this might be fixed at the MicroPython level now: https://github.com/micropython/micropython/commit/2b8de7436bbdb54da99bf6bdb79359876fd3bc1d

waffshappen:
> When using high ADC pin numbers, you now need to init `ADC(machine.Pin(29))` instead of `ADC(29)`
Yep, that fixes the battery voltage reading for me for now

PR #966, opened 2024-07-02
Examples: Update ADC()

helgibbons:
Should help with #965


helgibbons:
I've not done anything with our `Analog` class in `modules_py/pimoroni.py` here, which might also need looking at?

Gadgetoid:
Thank you!

I guess in Analog it should be something like:

```python
self.pin = ADC(pin) if isinstance(pin, Pin) else ADC(Pin(pin))
```

Looks like it's never actually called with a `Pin` instance, but it should probably handle that just in case.

Gadgetoid:
I think this has been rendered redundant by https://github.com/micropython/micropython/commit/2b8de7436bbdb54da99bf6bdb79359876fd3bc1d

PR #968, opened 2024-07-03
Move nano specs hack into our module cmake files.

Gadgetoid:
We've been maintaining a downstream patch to disable C++ exception handling, which gives us some headroom before our builds get too big and overwrite themselves at runtime.

Forgetting to apply this patch is a constant source of debugging frustration. It turns out it can just be added to our USER_C_MODULES cmake files instead, where it will implicitly be included when building against our modules.


Gadgetoid:
:crossed_fingers:  for CI pass.

It's messy, but a lot less hassle to maintain (since the patch breaks every time the MicroPython port CMakeLists changes) can't believe I didn't think of this sooner...

dpgeorge:
I much prefer it's done this way than permanently applying these settings into the main rp2 CMakeLists.txt file.

PR #975, opened 2024-07-27
fix: initialize the uninitialized

mariobodemann:
While building all demos I found that several colors are uninitialized, and where throwing an error when building.

Sample "error":

```
pimoroni-pico/examples/stellar_unicorn/stellar_lava_lamp.cpp: In function 'int main()':
pimoroni-pico/examples/stellar_unicorn/stellar_lava_lamp.cpp:128:27: error: 'bright_r' may be used uninitialized [-Werror=maybe-uninitialized]
  128 |           graphics.set_pen(bright_r, bright_g, bright_b);
      |           ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
pimoroni-pico/examples/stellar_unicorn/stellar_lava_lamp.cpp:113:13: note: 'bright_r' was declared here
  113 |     uint8_t bright_r, bright_g, bright_b;
      |             ^~~~~~~~
```

I could have added the build flag for dismissing that warning, but I felt more explicit is better.

Let me know what you think.

### DIFF
--- a/examples/cosmic_unicorn/cosmic_lava_lamp.cpp
+++ b/examples/cosmic_unicorn/cosmic_lava_lamp.cpp
@@ -106,11 +106,11 @@ int main() {
 
     // calculate dark, medium, and bright shades for rendering the 
     // lava
-    uint8_t dark_r, dark_g, dark_b;
+    uint8_t dark_r = 0, dark_g = 0, dark_b = 0;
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Thank you, more explicit is indeed better. I try to build with Werror and Wall where possible, but we’re still targeting a pretty old compiler 😬

PR #980, opened 2024-08-08
CI: Bump Pico SDK to 2.0.0. 🎉 

Gadgetoid:
Unleash the bugs of war!


Gadgetoid:
Note: There may be RP2350B (48 IO) bugs remaining!

PR #991, opened 2024-08-23
Update setup instructions for macOS

stephenhouser:
The preferred method for installing the ARM toolchain has changed. The method currently shown results in a compilation failure:

```
arm-none-eabi-gcc: fatal error: cannot read spec file 'nosys.specs': No such file or directory
```

Refer to [pico-sdk issue 1529](https://github.com/raspberrypi/pico-sdk/issues/1529)


Gadgetoid:
Thank you!!

Issue #1005, opened 2024-10-01
Pico2 slows down display pack


peardox:
Yeah, I couldn't believe it either...

Soldered up 2x Pico 2 and 1x Pimoroni Pico Plus 2 to try them out on diplsy packs

On my Pico Ws the display_pack_2_demo runs at 30fps (I added frame timing)

On all Pico 2 devices the FPS drops tro 24fps - tried all three devices I have to hand

Tested on a display pack 2.8 - I have the others as well but doubt there'll be much change (apart from the smallest as it has a smaller screen), the 2 + 2.8 are just a bigger LCD, exact same pixel counts

peardox:
Ahh - just spotted #985

@Gadgetoid I'll stick my code on Github in a little while and add a note here

peardox:
Here you go - https://github.com/peardox/display_pack_28

Gadgetoid:
I'm happy if you want to make a pull-request of it and claim the glory :laughing:

I had assumed you were using MicroPython, so I was waaaaaay off.

peardox:
I was already thinking of doing a PR with the modifications just for the display_pack_28 definitions but also adding the FPS demo will be nicer

I wanna check out the SPI rate as (150/133) * 62,500,000 |= 75,000,000 (actual value is a funny number so wanna experiment first)

peardox:
And yet changing it to...

```
//PicoGraphics_PenRGB332 graphics(st7789.width, st7789.height, nullptr);
PicoGraphics_PenRGB565 graphics(st7789.width, st7789.height, nullptr);

```

Dropped the frame rate from 45.x to 43.5-ish

peardox:
Oops 888 crashes - mem requirements are 240k for that res but I am on a Pico2 so plenty to spare

Looks like a different bug...

Slion:
> Dropped the frame rate from 45.x to 43.5-ish

That's weird, maybe there is something else at play. You better look at the frame time or even just the update time when trying to optimise your display performance. Though FPS is still interesting.

peardox:
Tried Pen4, Pen8, Pen332, and Pen565 which all work

Only Pen888 dies horribly (Pen4 just looks horrible)

Slion:
> Pens other than 565 all run thru a frame conversion function

Yup

peardox:
Now things get even weirder - switched to a PicoW and 332 is 30spf while 565 is 33fps (i.e. faster as you imagined it would)

[...]

Issue #985, opened 2024-08-13
Issue: 2350 possible SPI baud rates are different.


AndrewCapon:
Hi Guys,

This one may have some impact on all your SPI code.

I noticed it in the st7789 code, it was asking for baud 62'500'000 and getting 37'500'000.

I fixed it here with:

```
    // The ST7789 requires 16 ns between SPI rising edges.
    // 16 ns = 62,500,000 Hz
    // 2350 doesn't support 62,500,000 so use 75,000,000 seems to work.
#if !PICO_RP2350
    static const uint32_t SPI_BAUD = 62'500'000;
#else
    static const uint32_t SPI_BAUD = 75'000'000;
#endif
```

So just a heads up that you may be seeing some issues...

Cheers

Andy

peardox:
Well, that fixed my slow display :)

PR #1006, opened 2024-10-01
Add Pico Display Pack 2.8 and fix SPI for st7789 on Pico2

peardox:
This adds libraries/pico_display_28 along with a demo that displays the FPS in examples/pico_display_28

Also updated drivers/st7789 as per #985 / #1005

Full repo build impossible owing to hardware/rtc.h not existing for Pico2 when it hits breakout_encoder_wheel (about to raise issue) so just built and tested pico_display_28


Slion:
Why do we need a specifix example for 2.8? I thought is was exactly the same as the 2 but physically larger.

peardox:
> Why do we need a specific example for 2.8? I thought is was exactly the same as the 2 but physically larger.

It is't a carbon copy. It changes the text from "Hello World" to one that displays the Frames Per Second which is calculated on every update so has a benefit not present in other demos (as none of them do FPS)

Slion:
> The baud rate is dependant on the device in modified driver/st7789.hpp

Yeah I saw that I was just wondering if the faster baud rate also works on Pico 1.

peardox:
> Yeah I saw that I was just wondering if the faster baud rate also works on Pico 1.

I most sincerely doubt it (might give it a go tho)

The baud rate is tied to the system clock and those values are 1/2 device clock frequency (Pico 1 should be 66.5 for 1/2 freq - might be worth a try) although the docs say 16us apparently

Gadgetoid:
Sorry, it's been a busy time and I haven't got 'round to testing this yet 😬 

Nonetheless, this is appreciated thank you!

Slion:
In fact you really need that header defining the RGB pins cause they were changed for that display.

Gadgetoid:
Thank you!

PR #1011, opened 2024-10-07
PicoGraphics: Layers. (EXPERIMENTAL)

Gadgetoid:
Add preliminary support for multiple layered drawing surfaces.

Allows, for example, static content to be loaded into one layer and remain unmodified while the above layer contains animations.

Particularly useful for drawing PNG or JPEG UI elements which are then overdrawn with text or animated elements, without paying the cost of loading/decoding every frame.

Support for *two* layers is currently hard-coded for RGB332 and RGB565 pen types. There is *no* alpha blending, a pixel from the top layer is shown if it is not black (0) otherwise the underlying layer is displayed.

Since RAM usage is a multiple of layers, RGB332 will take 320*240*num_layers and RGB565 would be 320*240*2*num_layers, this is very much better suited for Pico 2 albeit RGB332 works well on a Pico.

Layer support is provided by the `frame_convert` function, as such it must *always* be called, even if the display native type matches the graphics buffer. EG: ST7735 and ST7789 displays supporting RGB565 now must call `frame_convert` even if your display buffer is RGB565. This could be improved by checking the number of layers and skipping the call if it's 1.

# TODO

1. [ ] Add support to all pen types
2. [x] Add fast-path to skip layer support when layers == 1 and buffer pen type == display pen type
3. [ ] Explore basic blending options, these will slow things down but could be cool
4. [x] Decide if we want to support *only* two layers, in which case `layers=` should be `layered=`
5. [ ] Benchmark, since we're now calculating an additional offset for every pixel drawn (could cache this when `set_layer`) is called?


Gadgetoid:
Decided on an arbitrary layer count, since PSRAM makes RGB565 possible with moar layers and... what else are we gonna use that RAM for?

Additionally, three layers is kinda the sweet spot- not two. That gives a layer for background, a layer for foreground and a middle-layer for redraws... useful for all sorts of things.

Benchmarking should, in theory, not be too important since layers can be turned off and the extra codepath effectively skipped.

PR #1018, opened 2024-10-25
CI: Switch to action-gh-release.

Gadgetoid:
Switch from the long deprecated upload-release-asset to action-gh-release and attempt to replicate the same behaviour.

Avoids nodejs and set-output deprecation warnings:
- https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/micropython.yml
+++ b/.github/workflows/micropython.yml
@@ -127,11 +127,6 @@ jobs:
 
     - name: Upload .uf2
       if: github.event_name == 'release'
-      uses: actions/upload-release-asset@v1
-      env:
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Note: I have created a test release to make sure this change actually works! We should be cautious during the next release just to be sure. I have put off this change for too long, it's time to pay back some:

✨  technical debt  ✨

PR #1019, opened 2024-10-25
PicoVector Enhancements & PicoGraphics Layers

Gadgetoid:
Merges and replaces #853 and #1011, since these features both relate to drawing and we usually need both of them in experimental builds.

⚠️  This is built against our `experimental/psram` branch of MicroPython which we ideally don't want in production. If/when we merge these features, this CI change should not be included!

# PicoVector

Incorporates the updated Pretty Poly library, which required a near complete rewrite of Pico Vector and will probably inform a redesign of the Python API.

There's now a proper C++ PicoVector library which will underpin the MicroPython bindings.

Alright Fonts are in an absolute state of turmoil, pending us figuring out how they might work. As the very least, the "af" format might change and break compatibility with the initial PicoVector release.

This is all a bit of a pain, but it will hopefully be worth it.

# Hub75 / Interstate75

Dropped RGB565 mode altogether, it should never be necessary since the displays are *very* low resolution anyway. Currently the largest supported display needs 64kb. If/when we add other modes, we should support them all and caveat that they will be much, much slower to flip.

Increased performance of RGB88 mode flip from 13.8ms down to about 2.87ms. This should make it much more feasible to hit 60FPS since user code costs additional time over these baseline numbers.

# Layers

Add preliminary support for multiple layered drawing surfaces.

Allows, for example, static content to be loaded into one layer and remain unmodified while the above layer contains animations.

Particularly useful for drawing PNG or JPEG UI elements which are then overdrawn with text or animated elements, without paying the cost of loading/decoding every frame.

Support for *two* layers is currently hard-coded for RGB332 and RGB565 pen types. There is *no* alpha blending, a pixel from the top layer is shown if it is not black (0) otherwise the underlying layer is displayed.

Since RAM usage is a multiple of layers, RGB332 will take 320*240*num_layers and RGB565 would be 320*240*2*num_layers, this is very much better suited for Pico 2 albeit RGB332 works well on a Pico.

Layer support is provided by the `frame_convert` function, as such it must *always* be called, even if the display native type matches the graphics buffer. EG: ST7735 and ST7789 displays supporting RGB565 now must call `frame_convert` even if your display buffer is RGB565. This could be improved by checking the number of layers and skipping the call if it's 1.

# TODO

1. [x] Add support to all pen types
2. [x] Add fast-path to skip layer support when layers == 1 and buffer pen type == display pen type
3. [ ] Explore basic blending options, these will slow things down but could be cool
4. [x] Decide if we want to support *only* two layers, in which case `layers=` should be `layered=`
5. [x] Benchmark, since we're now calculating an additional offset for every pixel drawn (could cache this when `set_layer`) is called?
6. [ ] Fix alpha blending for anti-aliasing, serious fringing issues. See i75 Duo's clock example for instance.


berlincount:
What's the status of this PR? I guess I'd want the Vector changes so I can play with the Tufty 2040 more, and I'd also love to try layers there - or maybe at least some kind of Double Buffering for moving text on top of a a PNG ...

Gadgetoid:
You can grab one of the test builds from here: https://github.com/pimoroni/pimoroni-pico/actions/runs/12372911360?pr=1019

There's also something resembling documentation here: https://github.com/pimoroni/presto/blob/main/docs/picovector.md

berlincount:
> You can grab one of the test builds from here: https://github.com/pimoroni/pimoroni-pico/actions/runs/12372911360?pr=1019
> 
> There's also something resembling documentation here: https://github.com/pimoroni/presto/blob/main/docs/picovector.md

ooooh there goes my sleep tonight 😆 thanks!

Gadgetoid:
> ooooh there goes my sleep tonight 😆 thanks!

Can relate!

PicoVector is still getting the rough edges bashed off, but it's close to release. Or, at least, I haven't run into a serious bug for a while. Any feedback would be welcome!

berlincount:
heh, let's see. probably something for 38c3 next week ... what I'd actually want is an SVG loader (for something minimal, though) - I actually need tufty as a badge, and want it to
be a bit fancy ...

Gadgetoid:
I've added some commits that bring consistent `MP_ERROR_TEXT` usage across our codebase, plus switching some things to `mp_raise_msg_varg` where appropriate, this is to prepare for the merge of https://github.com/micropython/micropython/pull/16920 which will enable error compression on the RP2 port. Without `MP_ERROR_TEXT` our builds will just fail.

PR #1022, opened 2024-10-27
fix: [url] placekitten.com is down, someone created a drop in replacement, placecats.com

SteveClement:


Gadgetoid:
Thank you!

I do wonder if we should rewrite these to do something else - more reliable - in the long term. I had been using placekitten for what seems like forever, and took it for granted!

PR #1027, opened 2024-11-17
PicoVector 2 text word wrap and transform

MichaelBell:
A couple of improvements to the PicoVector 2 text support:
- Add optional max line width (by doing word wrap) and max height arguments to text rendering.  These are in pixels, not scaled by font size, but they are scaled (and rotated) if the overall PicoVector transform is not the identity.
- Apply the overall PicoVector transform to text rendering.


examples/pico_display_2/pico_display_2_vector.cpp
```diff
@@ -51,7 +51,7 @@ int main() {
         pp_mat3_translate(&pos, 50, 50);
         pp_mat3_rotate(&pos, a);
         vector.draw(poly);
-        vector.text("Hello World", &pos);
+        vector.text("Hello World", 0, 0, &pos);
```
MichaelBell, line 54: Alternatively could make these optional arguments if you prefer - I had kind of forgotten that this was also exposed to C++.

examples/pico_w_explorer/pico_w_explorer_vector.cpp
```diff
@@ -41,8 +41,8 @@ int main() {
 
     vector.draw(poly);
 
-    //pp_mat3_t t = pp_mat3_identity();
-    //vector.text("Hello World", {0, 0}, &t);
+    pp_mat3_t t = pp_mat3_identity();
+    vector.text("Hello World", 0, 0, &t);
```
MichaelBell, line 45: Possibly uncommenting this was unwise as I don't have one of these to test.  I got confused between this and the new RP2350 based explorer 😆

examples/pico_display_2/pico_display_2_vector.cpp
Gadgetoid: I think it would be good to make these optional, since it also bubbles up to Python. C isn't super helpful about that, though. I wonder if it makes sense to have them part of the text metrics, or if that's just too conceptually weird.

Sensible API design for this has been quite tricky!

Gadgetoid:
Okay I think I merged this into #1019 and forgot. Closing for now and we can deal with any warts as they pop up.

PR #1033, opened 2024-12-04
hub75: don't enable an unrelated IRQ

Daft-Freak:
DREQs are not IRQs, `DREQ_PIO0_TXn` happens to align with `TIMER_IRQ_n` so this enabled a timer IRQ.

(This is something I happened to notice while poking around in there)


Daft-Freak:
More importantly, just noticed that stop would disable that IRQ which seems a lot more likely to break things...

Gadgetoid:
😬 I feel this pattern might be copied into a few places, will have a hunt. Thank you!

PR #1034, opened 2024-12-05
picographics: update text docs

Daft-Freak:
Spend a good few minutes trying to work out what `s` and `a` did.

"Is `s` scale?"
"Why it it 2 though?"
"`a` is... alpha? No, it defaults to 0..."
"Fine, I'll check the docs."
"That... did not help."

(This could've been avoided if they were called `scale` and `angle` instead...)

I didn't check the rest of the methods.

### DIFF
--- a/libraries/pico_graphics/README.md
+++ b/libraries/pico_graphics/README.md
@@ -304,14 +304,14 @@ PicoGraphics::circle(const Point &p, int32_t radius)
 ### Text
 
 ```c++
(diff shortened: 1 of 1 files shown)


Gadgetoid:
*saves milliseconds with shorter argument names*

*costs minutes in confusion and documentation trawling*

🫠 

Thanks, I wonder if we can deprecate and change these without great turmoil!

PR #1038, opened 2024-12-15
Hub75: Fix LED ghosting.

MichaelBell:
This adds a delay after latching the row before enabling the output, preventing the annoying ghosting of the next LED's colour onto the previous LED.


Gadgetoid:
Can confirm, ghosting-b-gone! Thank you.

PR #1043, opened 2024-12-26
Fix Button class on Pico 2W

isaac-webb:
I just picked up a Pico 2W and GFX Pack to dip my toes into Raspberry Pi microcontrollers. While experimenting, running the button demo didn't work. I stumbled upon a few forum threads and eventually discovered #1041 

This pull request modifies the `Button` class to specify the mode of the `Pin` object created within and, in my testing, this seems to fix the original example/built-in functionality that #1041 bypasses by locally creating `Pin` objects.

**Note:** I don't have the hardware to test this on any other Picos, so I only verified on my Pico 2W.

### DIFF
--- a/micropython/modules_py/pimoroni.py
+++ b/micropython/modules_py/pimoroni.py
@@ -96,7 +96,7 @@ class Button:
         self.invert = invert
         self.repeat_time = repeat_time
         self.hold_time = hold_time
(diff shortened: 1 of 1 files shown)


Gadgetoid:
I think this is a generally safe change to make, thank you!

(Hopefully not famous last words)

PR #1053, opened 2025-01-22
Plasma: Expose non-blocking update to MicroPython.

### LINKED ISSUE #1052: Expose blocking param of WS2812's update to micropython
I'm using the `update()` binding added for #236, in a loop running at my target frametime.

I'd like to get started on the next frame immediately after submitting the previous, but currently I believe it blocks until it's finished writing. 

I don't understand the threading/processing model in rp2350/2040, so this might not make any sense as a request, but my hope is that if the background repeating timer seems to be able to run without blocking python when it triggers, then perhaps `update(blocking=False)` might work as well.

For context, on my Plasma 2350W, with 300 LEDs configured, my (probably flawed) profiling says:
- 5ms minimum per frame if using `start()` only
- 16ms minimum per frame if set to RGBW, using `update()`
- 13ms minimum per frame if set to RGB, using `update()`

That's very close to my napkin calculation of the write time: 1.25us * 32 bits (RGBW) * 300 LEDs = 12ms, or 9ms for RGB.

Gadgetoid:
Add support for non-blocking updates to Plasma WS2812 and APA102.

Should fix #1052

TODO: Expose busy status, since the current setup simply ignores subsequent updates if DMA is already busy.

### DIFF
--- a/drivers/plasma/apa102.cpp
+++ b/drivers/plasma/apa102.cpp
@@ -60,14 +60,18 @@ bool APA102::dma_timer_callback(struct repeating_timer *t) {
     return true;
 }
 
+bool APA102::is_busy() {
+   return dma_channel_is_busy(dma_channel);
+}
+
 void APA102::update(bool blocking) {
-    if(dma_channel_is_busy(dma_channel) && !blocking) return;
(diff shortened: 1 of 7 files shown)


Gadgetoid:
Added to #1019

PR #1064, opened 2025-03-13
BME69X: Add C driver and MicroPython bindings.

Gadgetoid:
This is just a straight copy and paste of BME68X, dropping in the BME69X Sensor API and wishing upon a star that it works.


Gadgetoid:
Note to self: the builds here would not flash on my board. That suggests that the current main builds are non-functional 😬

helgibbons:
The Pico W build and the Enviro build are working OK for me, with:

```
"""BME69X demo

Demo of how to read the BME690 sensor.
"""

import time
from breakout_bme69x import BreakoutBME69X, STATUS_HEATER_STABLE

bme = BreakoutBME69X(machine.I2C(0), 0x76)
# If this gives an error, try the alternative address
# bme = BreakoutBME69X(machine.I2C(0), 0x77)

while True:
    temperature, pressure, humidity, gas, status, _, _ = bme.read()
    heater = "Stable" if status & STATUS_HEATER_STABLE else "Unstable"
    print("{:0.2f}c, {:0.2f}Pa, {:0.2f}%, {:0.2f} Ohms, Heater: {}".format(
        temperature, pressure, humidity, gas, heater))
    time.sleep(1.0)
    ```

PR #1067, opened 2025-03-21
plasma: Drop pin definitions from C module.

Gadgetoid:
Remove pin definitions in anticipation of them moving to pins.csv.

Eg:

```
from plasma2040 import BUTTON_A
machine.Pin(BUTTON_A)
```

Should become:

```
machine.Pin("BUTTON_A")
```

This is part of the Plasma migration: https://github.com/pimoroni/plasma/issues/1


Gadgetoid:
Added to #1019

PR #1072, opened 2025-04-13
Add support for unicode characters.

ihaanaho:
Add support for UTF-8 characters in picovector. The original implementation won't print the non-ASCII characters. 

The .af font file needs to have the necessary glyphs generated by afinate for the unicode characters to be shown.


Gadgetoid:
Nice! Very much for this as long as we can avoid tanking performance- I suspect any char lookup wrangling will be a fraction of the overall render time though.

I added a version flag to Alright Fonts recently when making some other changes, so we can potentially add a Unicode feature check to make sure the font actually supports it.

ihaanaho:
I don't have any objective means to measure performance, but subjectively I did not see any issues on Interstate75W. I do understand, however, that this implementation is shared by different kinds of HW.

FWIW, I'm using this in a scrolling text display and constantly doing the font file lookups (maybe not too ideal), and the .af file is also quite huge, 128 kB, containing characters way beyond my expected need.

libraries/pico_vector/alright-fonts.h
```diff
@@ -76,7 +76,7 @@ typedef struct {
 } af_path_t;
 
 typedef struct {
-  char codepoint;
+  int16_t codepoint;
```
Gadgetoid, line 79: uint16_t?

ihaanaho: Good catch, thanks. I'll fix this.

Gadgetoid:
Thank you, I have cherry-picked this on to https://github.com/pimoroni/pimoroni-pico/pull/1019 so I can get it into some test builds.

Gadgetoid:
Did you do anything special to afinate or otherwise to get unicode characters working?

I packed up some Material Icons (after some small modifications to coax Afinate into accepting a corpus of codepoints) and then attempted to display them with `vector.text("\ue951", 0, 0)`

```
>>> bytes("\ue951", "unicode")
b'\xee\xa5\x91'
```

The actual bytes produced seem reasonable and the unpacking in C recovers the correct codepoint:

```
>>> hex(((0xee << 12) | ((0xa5 & 0x3f) << 6) | (0x91 & 0x3f)) & 0xffff)
'0xe951'
```

But the glyph is not displayed!

A quick hack to afinate to move those codepoints down into ASCII gets them displaying fine.

Gadgetoid:
Aha thank you.

I’m running well outside that range so I’ll have to poke some more.

ihaanaho:
I see (\ue951), could it be that the source font file does not have those higher code point glyphs?

Gadgetoid:
It’s an icon font with about 2000 code points… remapping them down to the ASCII range works though that doesn’t prove I’m not messing up the conversion somewhere 🤣 codepoint map is here: https://github.com/google/material-design-icons/blob/master/font/MaterialIcons-Regular.codepoints

I should probably check in my hacks to afinate, but I mostly just made the corpus file reading support the code point list “format” in the file above.

P.S this change is actually merged into main and available in some of our builds! Thank you. More to come as I round-robin update everything.

Gadgetoid:
Got it! `find_glyph` was still `af_glyph_t *find_glyph(af_face_t *face, char c)` and needed to be `af_glyph_t *find_glyph(af_face_t *face, uint16_t c)`

ihaanaho:
Looks awesome! Good to hear you found the culprit 👍 Having the icons definitely brings a lot of new opportunities.

Glad I was able to contribute and give a tiny bit back.

PR #1080, opened 2025-05-28
Fix line break on text

### LINKED ISSUE #1076: PicoVector2 - Missing character on wrapped lines
This code
```
vector.set_font('Roboto-Medium.af', 24)
vector.text("This text is very long and should wrap around", 10, 50, max_width=220)
```
produces this result on the display:
```
This text is very long
nd should wrap 
round
```

Every first character on wrapped lines is missing.
Seems like a bug in af_render?

josch0:
This fix resolves #1076

Line breaks in multi-line texts are now handled correctly.


Gadgetoid:
Thank you, sorry I didn't get to your issue sooner- busy busy!

PR #1095, opened 2025-07-30
Update ci shell fix path

kisChang:
1. use absolute paths.
can to build my own firmware using this project with new code

2. cmake_configure add parameter input


Gadgetoid:
Thank you. I'm currently working on backporting our newer (badgeware) CI tooling here, which should completely eliminate this class of bug and make it runnable locally. Will land as part of #1139

PR #1129, opened 2026-04-28
Inky73: AC Waveform

thirdr:
- Adds additional command to the power off routine. 
- Compatible with the new AC Waveform panels


Gadgetoid:
Doo doo doot.

Issue #1135, opened 2026-08-05
RV3028: `BreakoutRTC: breakout not found when initialising`


helgibbons:
Breakouts from the latest batch show correctly up on I2C but fail when you try to init the breakout - this is because the chip ID register contains a different value (0x44 instead of 0x33).

We'll need to update the C++ driver to allow for this new chip ID.

helgibbons:
I did notice that our Raspberry Pi Python driver worked without modification, presumably because it doesn't check the chip ID - perhaps amending this driver to remove the check might be a more future-proof solution?

Gadgetoid:
I think in an awkward sense it's *useful* to know when the chip ID check fails- lest the chip itself actually have changed enough to be incompatible with the driver. Most drivers should enforce it, and - from recollection - tend to.

Also a bare minimum guard against the driver picking up some completely wrong chip and potentially doing something confusing or even damaging with it.

PR #1136, opened 2026-08-05
rv3028: add new chip ID/version

helgibbons:
should hopefully help with #1135


helgibbons:
Tested and working with old and new breakouts on Pico

Gadgetoid:
Thank you!

